PR #239, opened 2015-01-15
Update for many breaking changes.

alexchandel:
This addresses changes in closure, macro, slicing, and array syntax. This doesn't fix the build for me though, since I still see errors that `Decodable` and `Encodable` aren't implemented by `theme::Theme`, and that `Clone` isn't implemented by `widget::Widget`, even though all three of those traits are derived.


davekong:
I looked over this and it looks generally good to me; however, I wonder if the int and uint changes should be made in a separate PR and `#[feature(int_unit)]` should be added temporarily to lib.rs. 

Also, I don't really think that formatting changes unrelated to the fixes (found in the first commit) should be part of the PR.

For reference, The issue related to the ICE and broken Pi matrix widget in the demo is rust-lang/rust/issues/21216.

alexchandel:
I didn't see any code formatting changes. I tried to keep this a build-fix only.

Do you object to any of the uses of `usize` or `isize`? There are only about 30 uses altogether, so we can review it now. Looking at [the commit](https://github.com/alexchandel/conrod/commit/60e6f9d1578222d07b68e007c1763717d3b810d9) shows that they're all used for either indexing or lengths.

davekong:
I think this commit is important enough to be merged now and new issues can be opened if needed to address anything missing from the PR.

PR #243, opened 2015-01-19
Add quack dependency.

davekong:
I am not sure how this was compiling before since the all_widgets example
depends on quack, but I don't see a commit removing the quack dependency.


bvssvni:
I think this was a recent change in Cargo.

PR #279, opened 2015-01-28
Update to latest piston-quack

bvssvni:


eddyb:
Nice! You might want to come up with some methods to use a light accessor-like DSL to implement all of these easily.

PR #280, opened 2015-01-28
Added `#inline(always)` to property methods

bvssvni:


eddyb:
Yupp, definitely in desperate need of a macro.

bvssvni:
@eddyb Working on it!

PR #283, opened 2015-01-29
Use quack macro for label

bvssvni:


src/label.rs
```diff
@@ -81,65 +81,18 @@ impl<'a> Label<'a> {
...
-    }
-}
-
-impl<'a> SetAt for (Text<'a>, Label<'a>) {
-    #[inline(always)]
-    fn set_at(Text(text): Text<'a>, label: &mut Label<'a>) {
-        label.text = text;
-    }
+quack! {
+label: Label['a]
```
eddyb: I bet you can put `self` here and it will work.

bvssvni: Got

```
<quack macros>:8:27: 8:33 error: expected identifier, found keyword `self`
<quack macros>:8 $ val : $ set_prop_type , $ this : & mut $ this_type < $ ( $ t ) , * > ) {
                                           ^~~~~~
```

Issue #212, opened 2014-12-10
Separate `UiContext` from builders


bvssvni:
I am not sure whether ui_context is needed by the builders or not. It is probably used by some relative position methods. If we separate them, then controls can be declared before use, similarly to how piston-graphics does it.

bvssvni:
Done.

PR #286, opened 2015-01-29
Moved UiContext to draw method

bvssvni:
See https://github.com/PistonDevelopers/conrod/issues/212

* Merges `XContext` and `XBuilder` to `X`
* Makes it easier to add generic parameters on UiContext
* Takes UiContext as parameter for `up/down` methods


bvssvni:
I figured out it would be easier to start over with the current design. As you can see, the `up/down` methods is still there, but taking a `UiContext` as extra parameter.

One side benefit from this change is that widget can be declared before they are used.

PR #300, opened 2015-02-02
Fixed toggle matrix demo in all_widgets example

mitchmindtree:


eddyb:
Why is conrod using `Box::new` instead of `box`?

mitchmindtree:
No idea! I don't think any of these closures should be boxed anyway, looks like it might've been a "quick fix" that occurred when unboxed closures became the default? I might have a go at unboxing them now and see if I run into any issues.

bvssvni:
@eddyb I thought there was an stabilization issue with `box`.

eddyb:
@bvssvni `Box::new` is just there in case something goes horribly wrong, nobody using nightlies should touch it IMO.

PR #304, opened 2015-02-03
Unboxed all callback closures, closes #301

### LINKED ISSUE #301: Unbox the callback closures
As far as I can tell, it seems the callback closures were boxed as a sort of quick-fix during rustc's transition from boxed to unboxed closures. Please let me know if this is wrong and there's some other reason they are boxed, otherwise I'll start the change-over now :)

mitchmindtree:
Note that there is not yet a default type for the generic callback type parameter, meaning the user is forced to give a callback whether they want to or not. In most cases callbacks are given anyway, but for widgets like TextBox and EnvelopeEditor it isn't always necessary.

Perhaps we could create some empty Dummy type that impls the necessary FnMut for each widget as a default for the generic callback parameter so that the widget's type is always satisfied whether or not the user gives a callback?

@bvssvni @eddyb are there nicer ways to handle this?

### DIFF
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -177,7 +177,7 @@ fn draw_ui(gl: &mut Gl,
             .rgba(0.4, 0.75, 0.6, 1.0)
             .frame(demo.frame_width)
             .label("PRESS")
-            .callback(box |&mut:| demo.bg_color = Color::random())
+            .callback(|&mut:| demo.bg_color = Color::random())
             .draw(uic, gl);
 
     }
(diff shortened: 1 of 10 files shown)


bvssvni:
I think it should be possible to use `F = ()` as default type since the `where` constraint is only enforced when the callback is set on the widget.

mitchmindtree:
oh of course!

:+1:

I'll add it now

bvssvni:
I opened https://github.com/PistonDevelopers/conrod/issues/305

Issue #317, opened 2015-02-15
7guis


bvssvni:
https://github.com/eugenkiss/7guis/wiki

* 7 small challenges as "usability benchmarks" for GUI library design

mitchmindtree:
Wow these are great! We have a lot of work to do :wink:

Issue #321, opened 2015-02-21
The Road to Pure Rust


mitchmindtree:
Requiring users to compile / install C libraries is a real pity. It would be amazing if Conrod required nothing more than
```
[dependencies]
conrod = "*"
```
I think we still have two dependencies that require having C libraries installed on the user's system:

- [x] SDL2

~~This is only used for the example, but should be easily replaced by glutin :) #312~~ Conrod is now backend agnostic and all examples now use glutin.

- [ ] FreeType

This one may be trickier to handle. Is anyone aware of any purely rust font-rendering?

bvssvni:
Text is basically vector shapes, and Valve has a paper on that http://www.valvesoftware.com/publications/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf

jhasse:
@bvssvni The paper is only about rendering vector shapes, not about loading font files, which is also a big part of FreeType's job.

I don't think FreeType can ever be replaced. It should rather just ship with Rust (on Windows / OS X) so that users don't need to install it.

viperscape:
svg has a font and text spec. I was actually looking into this, but realized how complex svg is, even "tiny" spec. So I've been wondering about an xsd parser, to parse the data structure in to rust. Any ways, svg might be an alternative to consider

viperscape:
relevant [svg font spec](http://www.w3.org/TR/SVG/fonts.html)
[example svg font](https://github.com/adobe-fonts/source-code-pro/releases/tag/1.017R)

bstrie:
Could a cargo-feature be used to completely disable the need for freetype? For example, many of the examples in the piston-examples repo don't need freetype for anything, but you still must manually install it in order to compile the examples, which is quite annoying.

bstrie:
> It should rather just ship with Rust (on Windows / OS X) so that users don't need to install it.

@jhasse, are you suggesting that the Rust developers should ship freetype binaries in the standard Rust distributions? Because that seems very out of scope for the Rust project.

jhasse:
Yes, I'm suggesting that. Why is it out of scope?

bstrie:
Rust is a general-purpose systems programming language and freetype is a library for font rendering. It's not Rust's responsibility to provide a font rendering library on platforms where freetype does not exist.

jhasse:
It's already shipping a few MinGW libraries though, but you're right: It shouldn't ship by default.

[...]

PR #325, opened 2015-02-22
Changed all_widgets over to glutin (still needs some fixes), started basic counter example

mitchmindtree:
Not sure if we should merge this yet, however at least the tests would pass as sdl2 seems to still be broken.

Remaining issues:

- the widgets in the "all_widgets" example seem to render a few pixels higher than they should (if someone could test this on windows or linux that would be great).
- everything halves in size on retina display. I think this is because unlike `sdl2`, `glutin` works with a literal pixel count (2.5k+), so maybe we should have an option to halve the res with `glutin_window`? Maybe glutin already offers this though? @bvssvni

I think it'd be cool to try and implement each of the examples from here #317 as nicely as possible as conrod could use some simpler examples and it would also encourage us to address some fixes and new features.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,28 @@
 language: rust
+
 notifications:
     irc: "irc.mozilla.org#piston-internals"
+
 os:
     - linux
     - osx
+
 env:
     global:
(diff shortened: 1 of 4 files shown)


Potpourri:
On linux are identical:
Glutin
glutin
SDL2
sdl2

mitchmindtree:
@Potpourri great! I might merge this for now and keep an eye out for a fix for the strange osx behaviour. #321

bvssvni:
@mitchmindtree I'll open up an issue about retina.

PR #338, opened 2015-03-23
Updated to latest io, made cargo always use latest deps, changed over to slice syntax

mitchmindtree:


bvssvni:
Nice!

mitchmindtree:
It seems to be the deps that are failing travis, pretty sure it should be alright to merge!

Issue #341, opened 2015-03-28
TextBox cursor can break


davekong:
The TextBox widget seems to assume that the String it modifies is only modified by it and lives for the lifetime from the first creation of the widget to the last even though this is not enforced. Create a text box and pass in a new string to TextBox::new() every time the widgets is drawn and two things will happen.

1. If the cursor is advanced two characters past the end of the string, the program will panic and crash.
2. If a character is entered that is a different width than the character that will be in front of it the next time the box is rendered, the cursor will advance into the middle of the next character or somewhere past the end of it.

One idea to solve this issue is to give ownership of the String to the text box. My understanding is that this would be associated with the UIID of the text box and stored in a UiContext. So there would also need to be a method to get the string from the UiContext and a way to set the String without interacting with the text box which would properly reset the cursor.

mitchmindtree:
Thanks for the issue! I wrote TextBox in my early days of rust and it hasn't had much testing, so I appreciate it. Been meaning to come back to conrod and comb over everything, but I've been so busy with a personal project.
> The TextBox widget seems to assume that the String it modifies is only modified by it and lives for the lifetime from the first creation of the widget to the last even though this is not enforced.

I'm not quite sure what you mean by this. Are you trying to store the widget by any chance? (I'm not entirely sure if you know this already, but I just thought I'd check) Each widget in Conrod is designed to work in an "immediate mode" style (i.e. it behaves more like a function than an object - there's almost never a reason why you'd want to store a Conrod widget). [Here's an example of a typical `TextBox` usage](https://github.com/PistonDevelopers/conrod/blob/master/examples/all_widgets.rs#L393). You shouldn't have to worry about storing it or maintaining it's state (it will store its own state within the UiContext automatically and retrieve it each time it's constructed by using its unique ID). You may know all this already but just thought I'd clarify.

> 1. If the cursor is advanced two characters past the end of the string, the program will panic and crash.
> 2. If a character is entered that is a different width than the character that will be in front of it the next time the box is rendered, the cursor will advance into the middle of the next character or somewhere past the end of it.

These sound concerning :worried: The code for cursor movement [is here] if you want to have a look - I'm a few days behind the nightly atm and am unable to build conrod, but I'll have a look soon once I catch up.

davekong:
[...]

PR #345, opened 2015-03-30
Fix textbox cursor

davekong:
Fixes issue #341.

Note that the fix allows the cursor to go past the end of the TextBox up to the end of the text if the text is longer than the box.


bvssvni:
Thanks!

Issue #71, opened 2014-08-13
TextBox widget


bvssvni:
The state stores cursor position and selected range.

The text is passed in as argument and the new text is passed back to the callback.

davekong:
When the textbox is not captured there is no selection / cursor so the view can not be based on it. Therefore, I think it makes sense to set the view to whatever it was when the textbox was last captured, but what should happen if the text in the textbox is updated without being captured again first? Should the view reset to display whatever fits starting from the beginning of the text or slide left to the end of the new text or something else?

Currently, when the text is changed without interaction with the textbox and the textbox is captured the cursor / selection.end slides left if the text became shorter than the selection, otherwise it is unchanged.

davekong:
It looks like input boxes in the browser reset and display from the start of the text when capture is lost, so unless anyone objects, I will go with this for now. This also means I don't need to keep track of what the view was before capture was lost, which is nice.

bvssvni:
@dcampbell24 That's a good point! If this becomes annoying, then we'll change it.

TheNeikos:
If noone is working on the keyboard events I would like to take a shot at them.

dcampbell24:
@TheNeikos go for it. I can assign this to you if you want. I don't think I am going to be doing much with it any time soon.

TheNeikos:
So, interestingly [`UserInput`](http://docs.piston.rs/conrod/conrod/struct.UserInput.html) only stores keys since last input, and pressing both at the same time tick is not very usable. Before I start adding that state to the textbox is that intended behaviour in the long term? Because an array of pressed keys *right now* seems to be much better to work with.

michaeljones:
Trying to find the best place to ask this without opening a new issue. Is it possible to set the focus to a TextBox when it is first created? I'm creating a new one in a kind of pop up and would love it for to capture keyboard input straight away so the user can start typing without having to click on it with the mouse.

PR #351, opened 2015-04-03
Textbox select

davekong:
Adds support for selecting text with the left mouse button in a TextBox. Related functionality to follow soon.

cc: #71

### DIFF
--- a/src/rectangle.rs
+++ b/src/rectangle.rs
@@ -12,7 +12,7 @@ use ui_context::UiContext;
 use utils::map_range;
 
 /// Represents the state of the Button widget.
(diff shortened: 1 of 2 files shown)


davekong:
Some feedback is welcome before I add the rest of the selection related features from #71.

bvssvni:
Looks good.

Rust is going through a lot of breaking changes at the moment, so if we merge this now, the related functionality can be added in another PR with less chaos.

davekong:
OK, sure. I'll wait for the dust to settle before adding the rest.

davekong:
Will add the other changes later in a new PR after conrod is building and the beta is out.

Issue #220, opened 2014-12-14
Fixed-width textboxes that scroll


abonander:
This is relatively trivial to implement if the user provides the maximum column (character) count to fit in the textbox, then internally use a subslice that shifts its start/end if the user tries to push the cursor further left or right.

davekong:
Related: #71

PR #358, opened 2015-04-06
Add some more TextBox features

dcampbell24:
This first commit adds TextBox support for backspace and insert with a selection, but we probably don't want to merge yet since this update also allows text to be input past the end of the box in anticipation of the View features I will add next to cause a slid view instead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,4 @@
 #![deny(missing_copy_implementations)]
-#![feature(str_char)]
 
 #[macro_use] extern crate bitflags;
 extern crate clock_ticks;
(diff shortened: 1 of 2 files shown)


davekong:
Related to #71, #220.

dcampbell24:
I am waiting on making more changes because I want to use the new Rect type I proposed for graphics to make the drawing calculations easier to write and understand.

See https://github.com/PistonDevelopers/graphics/issues/882

mitchmindtree:
@dcampbell24 I'm going to merge this for now as I'm about to do a little module restructuring and I don't want to cause any conflicts - I don't think the text overflow is too bad a problem as it will only be temporary and it doesn't drastically affect conrod's usability :) If anyone has any qualms, feel free to un-merge!

PR #362, opened 2015-04-08
Upgrade examples to latest event

Potpourri:


bvssvni:
@Potpourri You were 30 second too late!

PR #368, opened 2015-04-15
Consolidated all widget types into widget module, cleaned and reformed color handling throughout crate.

mitchmindtree:
Quite happy with how the new color module has turned out! The crate also feels a little cleaner now the widgets are separated into a specific model.

Will continue to tweak at conrod over the next couple days... there are lots of small things I want to clean and fix, and a few issues I'd like to finally get around to tackling :)


bvssvni:
Nice!

PR #379, opened 2015-04-17
Text box view

dcampbell24:
When support for scissoring is added, just need to scissor the view when the text is drawn.

cc #71

### DIFF
--- a/src/widget/text_box.rs
+++ b/src/widget/text_box.rs
@@ -395,28 +395,47 @@ impl<'a, F> ::draw::Drawable for TextBox<'a, F>
         rectangle::draw(ui.win_w, ui.win_h, graphics, new_state.as_rectangle_state(),
                         self.pos, self.dim, maybe_frame, color);
 
-        if let State::Capturing(captured) = new_state {
-            let mut cursor = captured.cursor;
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Nice @dcampbell24 ! Thanks for all your awesome work with TextBox! It's definitely starting to look in far better shape :)

Travis seems fine, going to merge this now :+1:

bvssvni:
Scissoring works with gfx_graphics now. Take a look at the draw_state_test.rs example. If we port the Conrod example to piston_window then we can try it out.

dcampbell24:
@bvssvni I tried porting the counter.rs example over to piston_window and ran into two issues:

1. ui::handle_event expects the event to implement Debug, but PistonWindow does not implement Debug.
2. The Debug constraint can be removed and counter.rs will compile, but no text is drawn to the screen.

See https://github.com/dcampbell24/conrod/tree/counter-piston_window

bvssvni:
@dcampbell24 You need to call `GlyphCache::update` https://github.com/PistonDevelopers/gfx_graphics/blob/master/src/glyph.rs#L72 to create the font textures.

Issue #380, opened 2015-04-17
Layout / Positioning API Draft


mitchmindtree:
Discussion started [here](https://github.com/PistonDevelopers/conrod/issues/374#issuecomment-93914098).

Summary
-------------
Conrod should require that a user **never** has to think about or use the coordinate system. If a use case is found in which the user finds they have no choice but to use coordinates directly, then we should treat this as a serious API design flaw and address it.

To achieve this, we need an expressive and thorough API. I propose the following design.

Canvases (i.e. splitting up the window into sections)
-------------------------------------------------------------------
This should be the first step in the user's approach to widget layout.

In almost all cases, you want to be able to divide the screen into different sections (i.e. columns and rows) and have a custom name you can use to describe each section (i.e. `header`, `mid_left_column`).

To do this, I propose using something like @eddyb's super simple `vbox`/`hbox` (short for "vertical box" and "horizontal box") design.

Here are some pics showing eddyb's implementation [1](https://i.imgur.com/btcNQsV.png), [2](https://i.imgur.com/btbVGSw.png) and [3](https://i.imgur.com/WfJMqDW.png).

We could provide `vbox!` and `hbox!` macros to make this simple and easy.

For example:

```Rust
vbox![header, hbox![left_col, mid_col, right_col], footer];
```

would translate to 5 canvases:
- A `Canvas` called `header` occupying the top third of the screen.
- Three column `Canvas`es occuping the centre third of the screen named `left_col`, `mid_col` and `right_col`.
- A `Canvas` called `footer` occupying the bottom third of the screen.

The `Ui`s active `Theme` should specify default margins and padding, however we should also allow this to be described in the `vbox!` and `hbox!` macros, i.e.:

```Rust
vbox![
    header: top_padding = 20,
    hbox![left_col: left_margin = 30; top_padding = 30,
          mid_col: left_margin = 30; top_padding = 30,
          right_col: left_margin = 30; top_padding = 30],
    footer: bottom_padding = 50,
];
```

So far we're able to split up a screen into even sections, however a user may sometimes want to specify non-even splits with specific percentages of the screen. We could also describe this within our `vbox!` and `hbox!` macros:

```Rust
vbox![
    header: split = 20%,
    hbox![left_col: split = 50%, mid_col, right_col]: split = 60%,
    footer,
];
```

This would would result in:
- the `header` `Canvas` occupying the top 20% of the window.
[...]

Issue #374, opened 2015-04-16
Change to using a Cartesian plane for Conrod's positioning with (0, 0) at centre.


mitchmindtree:
As we'll likely use elmesque (which uses the Cartesian plane with centred (0, 0)) for graphics layout and positioning, Conrod should too. It is arguably a more generally intuitive co-ordinate system, and it also makes it easier to reason about elements' relative positioning.

bvssvni:
The upper left corner with y axis down is very common, also used by Window forms. This is what piston-graphics uses. Centered coordinate system is more common for 3D.

mitchmindtree:
Indeed, it is quite common, however using a cartesian plane and treating the origin as (0, 0) can offer some really nice properties for grouping elements and layout - Elm is a language designed for web GUI that also uses the cartesian plane, they describe these nice properties in [this post](http://elm-lang.org/blog/announce/0.8.elm) (under the "Better 2D Graphics" section). I'm halfway into #372 (I'm addressing this at the same time as #372 requires elmesque) and it's proving to be quite useful, but we can change it back if it causes any issues. The user will most likely just be using relative positioning (I'm fixing that atm also) most of the time so it probably won't make a huge difference from an API standpoint.

bvssvni:
Keeping a consistent absolute coordinate system is more important than using the same axis that people learn in math classes. This is because the majority of programmers will expect the coordinate system to start in the upper left corner, and it is important to make libraries work together.

bvssvni:
There are reasons why 2D graphics usually starts a coordinate systems from a corner. In relative coordinates, you need to know the size of the parent rectangle. If you start from the upper left corner, then you don't need that extra information. It also fits how you store data in a raw image format, with the first byte corresponding to the upper left pixel. There is a historical reason of how computer screens worked that is the origin of this coordinate system.

Cartesian coordinates does not specify that the y axis should point up, they are general for all choices of axis directions. When working with rotations, it is common to use y-axis up, but in widget programming rotations are less common and the fact that you don't have to know the parent rectangle or window size makes it practical. Some GUI systems use the lower left corner as origin.

I agree with the argument that it is easy to do mistakes when working with polar coordinates, but the other arguments seems fishy.

bvssvni:
[...]

Issue #373, opened 2015-04-15
Move `color.rs` module into its own crate?


mitchmindtree:
It's just a thought - I realise crates often need their own versions of the `Color` type depending on the context, however as there isn't currently a `PistonDevelopers/color` crate at all, perhaps conrod's color.rs would make a handy, general-use color crate? Perhaps we could merge the `read_color` and `select_color` crates into it also?

bvssvni:
`select_color` depends on the image library, but `read_color` could be merged into it.

mitchmindtree:
Closed by #385

Issue #354, opened 2015-04-04
Make `Positionable` inherit `Shapeable`


bvssvni:
The only widget that has position but no size is `Label`.

bvssvni:
I see some alternatives:

1. Assume that the size of the label is zero
2. Compute the size it uses
3. Add a size to the label
4. Add an optional size the label, which becomes either 1) or 2) when not specified

davekong:
I think having the size of the label is pretty useful. For example, I have a table (WidgetMatrix) and want to place a label above it. Based on #219 it sounds like 4 is a desirable choice with fallback to 2.

mitchmindtree:
Closed by #385

PR #385, opened 2015-04-20
BREAKING CHANGES - Library Reform (see comment for details).

### LINKED ISSUE #372: Defer "draw" until end of the render cycle
At the moment the "draw" stage is heavily intertwined with the "input -> callback" stage. The benefits of untangling them into two distinct stages are as follows:
- Draw only needs to be called on the `Ui` at the end of the cycle (after widget construction), i.e. `ui.draw(graphics, glyph_cache)`
- As a result we could properly handle draw ordering (i.e. we'll be able to draw open DropDownList's last).

Rather than calling `.draw(graphics)` at the end of each widget's construction, the `.callback(..)` method would indicate that the widget construction is complete and store the necessary drawing information in the `Ui` (In the case that the widget doesn't have a callback, we could include a `.set()` method which would indicate the same thing).

The necessary drawing information to store for each widget would simply be a graphics Context (just a collection of Lines/Rectangles in most cases).

mitchmindtree:
Sorry for the massive commit! The reform involved a lot of changes and it made sense to address some older issues while I was going. In general, I'm super excited about these changes :)

Here's a quick summary of changes before going into deets for a couple:

- Removed Drawable trait in favour of new `.set` method for deferred drawing (see below).
- Removed Callable trait and moved callback method to the widget impls to save on imports.
- Removed color.rs module in favour of PistonDevelopers/color crate.
- Fixed / vastly improved Positionable trait and removed annoying Ui parameter.
- Added methods for alignment (top, bottom, left, right, middle) i.e. `.down(20.0).align_middle()`. (the default vertical alignment is `Left` and the default horizontal alignment is `Top`.
- Switched `graphics` over to `elmesque`.
- Fixed issue where text would flow off widget.
- Deleted unnecessary `dimensions.rs`, `rectangle.rs`, `shape.rs`, `draw.rs`, and `point.rs` modules.
- Added a `Widget` struct to make the `Ui`s `widget_cache` much cleaner.
- Inlined a few unnecessary widget module functions that were only being called in one place.
- Changed the macro-generated `set_state` function to `ui.set_widget(ui_id, Widget {...})`.

Removed Drawable trait in favour of `.set` method.
--------------------------------------------------

This change allows us to store renderable Elements in the Ui for deferred drawing, rather than drawing the widgets at their call site. This will allow ourselves and users to set custom rendering depths for widgets and will also make it extremely easy to make sure that captured widgets are rendered last (i.e. drawing an open DropDownList on top of widgets that are called after it).

Also, I've moved the UiId from the widget constructor to this set method so the user only has to deal with `Ui` related things in a single method. i.e. this
```Rust
Button::new(ui_id).draw(ui, gl);
```
is now this
```Rust
Button::new().set(ui_id, ui);
```
Now, draw is only called once: on the `Ui` type after all the widgets have been set. When called, it will sort the widgets by rendering depth and then render all of the stored elements.
```Rust
ui.draw(gl);
```

Switched `graphics` over to `elmesque`
----------------------------------------------------

This is the change that made deferring the rendering of widgets possible - rather than having to directly draw the widgets at the widgets' call site, we can use elmesque to compose an `Element` and store it within our WidgetCache. Note that `elmesque` still uses the `graphics` crate under the hood, but offers a high level interface for composition and layout. One issue with this is that it will require more allocations in order to be able to store the data within the `Element`s and in turn this may impact performance. I have a few ideas for addressing this, but that will be for a future PR.

What next?
----------------

These changes open up a few doorways! These are what I'm hoping to implement over the next couple days:
- Register captured widgets within the `Ui` for consideration when sorting for rendering order.
- Scroll functionality for DropDownList.
- The Positionable trait is ready for easy integration with `Canvas`es (see #380), which will be one of the final steps towards offering a fully featured layout API without the need for coords :)
- Unique per-widget theme-ing.

Closes #372, #350, #354, #374, #373, #347, #349, #218, #344 and #145  :tada:


bvssvni:
Nice! Now we can use `Button` as a object-as-function just like in piston-graphics!

mitchmindtree:
@bvssvni has reviewed - there are a few things to address but they can easily be done in a future PR - going to merge :+1:

PR #386, opened 2015-04-20
draw the TextBox cursor like we used to

dcampbell24:
I am assuming we don't want a pulsating selection even though it may be fun at disco parties. Also, don't draw the selection over the frame.


mitchmindtree:
ahh sorry about that! I tried to not break too much stuff in text_box.rs for you - I hope it's not too bad!

dcampbell24:
No worries. It looks good. Thanks for all of the recent improvements!

PR #392, opened 2015-04-21
Added support for Custom widgets

mitchmindtree:


bvssvni:
Hmm... I don't like the generic parameter approach, because if two people Alice and Bob work on their custom widgets without knowing about each other, they have to change their code afterwards to agreed on the same state. Is there a way we can use `Any` to store a dynamic typed state here?

mitchmindtree:
@bvssvni I imagined that Bob and Alice, when they decide to work together and join their code, could wrap their different types as variants of a single enum? They only have to implement two methods for it, which would just call the same methods on the wrapped types?

I'm open to other approaches though :) I don't have much experience with `Any` but would be interested in seeing a draft on how it could be used. Perhaps we could merge this to add at least some support in the mean time?

libfud:
I'm definitely interested in using this.

mitchmindtree:
Alright, I'll merge this now and if we come up with something better in the future then we can replace it then :+1:

bvssvni:
@mitchmindtree I'll write up the idea.

bvssvni:
I opened https://github.com/PistonDevelopers/conrod/issues/397

Issue #397, opened 2015-04-21
Make custom widgets store `Box<Any>`


bvssvni:
Each widgets has a unique id, and widgets are not created and destroyed frequently. One can assume that the widget is the only object that needs to know the type of the state. So, instead of having a generic parameter that requires you to know the set of types upfront, the idea is that custom widgets stores their state in a `Box<Any>` which can store any data.

This solves a problem when you want to combine libraries with custom widgets.

bvssvni:
This is done.

PR #409, opened 2015-05-02
Made custom widgets use `Box<Any>`

bvssvni:
Closes https://github.com/PistonDevelopers/conrod/issues/397

* Removed `State` trait
* Removed `Custom::State`
* Removed `Copy/Clone` on `Kind/Widget`
* Added `#![feature(core)]` to check type id, will be removed when
`Any` gets associated static

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.0.9"
+version = "0.0.10"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,10 +1,11 @@
-//! 
+//!
 //! # Conrod
 //!
 //! An easy-to-use, immediate-mode, 2D GUI library featuring a range of useful widgets.
 //!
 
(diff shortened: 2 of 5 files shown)


bvssvni:
cc @mitchmindtree

mitchmindtree:
Nice! Looks like this will at least keep the library simpler if anything.

Just a note: Using `.get_type_id()` (or an associated static/const) instead of providing a `.matches` method doesn't allow a user to specify their custom widget kinds as variants of an enum? I guess this is ok, but maybe we should let the user know somehow (or they might run into strange errors)? 

Also, do you mind if we wait until the associated statics/consts become stable before merging this? I don't think this change is worth making Conrod unstable.

bvssvni:
We could use a `&'static str` for the widget kind use it for `.matches` with a recommendation of using `lib/type` format.

mitchmindtree:
@bvssvni Ok I have some ideas for changing over to both `Any` and a `Widget` trait, which will:
- remove the separation between both custom and internal types
- provide a super clear API for designing widgets
- automatically only update a widget's `Element` in the occasion that it's `State` or `Style` has changed

Working on it now, hopefully i'll get a PR in for reviewing by tonight :+1:

bvssvni:
Nice! I'll close this.

PR #410, opened 2015-05-02
Implemented the Split Canvas type for splitting a window into sections for layout.

### LINKED ISSUE #404: A good name for canvas positioning enum with variants such as `TopLeft`, `BottomRight`, `Middle`
There'll be an enum for placing `Widget`s upon a `Canvas`, I'm currently tossing up between `Location` and `Place` (I'd use `Position`, but we're already [using that](https://github.com/PistonDevelopers/conrod/blob/master/src/position.rs#L17)).

```Rust
pub enum $Name {
    Middle,
    TopLeft,
    TopRight,
    BottomLeft,
    BottomRight,
    MidTop,
    MidBottom,
    MidLeft,
    MidRight,
}
```

I'm leaning towards `Place` because it feels a little nicer to say in context - it's like saying I have `Widget` and I `Place` it on this `Canvas`, i.e.:

```Rust
Label::new("Fancy Title").place(Middle).canvas(HEADER).set(TITLE, ui);
```

I'm open to any other suggestions though :)

Edit: actually I'll probably also add a method for each variant to the `Positionable` trait so that the user doesn't have to import the Variants into the namespace, so the above example could actually be:
```Rust
Label::new("Fancy Title").middle_of(HEADER).set(TITLE, ui);
```

Note that if a canvas isn't specified, the `Ui` will just use the current canvas which can be set manually or will just use the last specified canvas.

mitchmindtree:
Also added an example, here's what it looks like (no co-ords used)!

canvasdemoscreenshot

There are also a new series of methods in the Positionable trait for placing widgets on canvasses.

There's also now a method on the `Ui` which can give you the dimensions of a canvas from it's `Id`, making it easy to set the dimensions of widgets to match.

The example uses the `piston_window` crate along with `glutin_window` - very nice and easy to set up! The only issue I have is that the resolution is halved on my retina display and the graphics backend doesn't seem to respond to the window resizing (I'll do up a separate issue).

Still need to:
- Make an optional method that makes the splits adjustable within some bounds.
- Implement a macro to allow for constructing a more efficient static Canvas tree (it uses dynamic allocation atm).
- Implement a draggable, sizeable `Floating` canvas kind for pop-ups and hovering windows.

Closes #404 and #380


mitchmindtree:
Going to merge this as travis is passing and cargo test passes on my end :) Lemme know if anyone has any qualms :+1:

bvssvni:
Awesome!

PR #419, opened 2015-05-05
Reserve a slot for Label before calling update_widget()

krdln:
Call `get_widget_mut()` before `update_widget()` to handle cases where Label would get out-of-bound index in widget cache.


mitchmindtree:
@krdln ahh thanks for this! I have some fairly big changes to the way widgets are being cached within the `Ui` that has removed the need for `get_widget_mut` altogether - they'll probably land tonight, so I'll leave this for now.

PR #420, opened 2015-05-05
Moves widget_cache over to `Any` and new `Widget` trait, removes unnecessary `CustomWidget` trait.

### LINKED ISSUE #401: Only construct new `Element` for a widget if state != new_state.
Now that `Element`'s `draw` method borrows the `Element` (instead of moving it), we should ensure that the `Element` used for drawing is only updated in the case that the Widget's state has changed. I'll do this now.

### LINKED ISSUE #402: Consolidate all styling required for drawing each widget into WidgetStyle types.
This will make it easier to address both #401 and #140. It will also make it easier to consistently separate functionality needed for drawing from widget logic.

mitchmindtree:
**DO NOT MERGE THIS YET**

A follow up on #409.

The widget cache no longer uses the enum representation of widgets and instead uses the `Any` trait for dynamically typing widgets.

A new `Widget` trait exposes a clearer, simpler interface and removes the distinction between custom widgets and internal widgets - it should now be *much* easier for external libraries/users to design their own widgets.

The new `Widget` trait also explicitly separates `State` updating from renderable `Element` construction. A nice property of this distinction is that it is possible for the `Widget` trait to automatically determine whether or not a new `Element` needs to be constructed for a `Widget` by requiring each widget's `State` and `Style` to implement `PartialEq` and comparing the previous to the new (this is what I need to finish implementing before this can be merged). This should significantly reduce allocations and increase performance considerably. This closes #401.

Note that this change now requires that the `Widget` trait be imported in order to use the `.set` method, whereas previously `.set` was implemented directly onto each widget.

Oh also, each widget now has optional, unique styling within the `Theme`! This closes #402.

This also closes #419 as it is no longer necessary.


stjahns:
Nice! Looking forward to this :)

bvssvni:
Nice! Looks good.

viperscape:
:+1:

bvssvni:
:tada:

mitchmindtree:
:dancer: 

@stjahns @viperscape let me know if there are any issues with the changes! Hopefully the `Widget` trait should make things a bit clearer if anything. I have [some ideas](https://github.com/PistonDevelopers/conrod/issues/424) for improving it further, but I'm not sure exactly when I'll get a chance as there are other priorities for the moment (i.e. floating canvasses, scrollability, etc)

PR #422, opened 2015-05-07
Reorganized assets (fonts & images)

debloper:
Also, nodalUIdemo.png can probably be safely removed from the repo (unless ref'd from elsewhere).


mitchmindtree:
Thanks for this :)

If you could move the Cargo.png back into the top level that'd be appreciated, as it's not really a conrod asset per say, just a picture of conrod's dependency graph :+1:

And yep, nodalUIdemo.png can definitely be removed.

bvssvni:
@debloper Can you fix the changes mentioned above so we can merge this?

debloper:
Added a new commit, instead of sqashed rebasing - hope that's okay. Wasn't sure which would be more expected from the contrib-doc.

Sorry for the delay, somehow this fell through the cracks... should be good to merge now. :D

bvssvni:
@debloper No problem!

Issue #390, opened 2015-04-21
Make prints of errors print out to stderr


bvssvni:
Printing to standard output messes up the application captured output in console environments, and therefore I think we should print the errors to stderr.

mitchmindtree:
Closed by #428

PR #428, opened 2015-05-08
Update the println!()s to print to stderr

Esption:
For https://github.com/PistonDevelopers/conrod/issues/390

Changed println!()'s to writeln!()'s to stderr.

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -19,6 +19,7 @@ use position::{Depth, Dimensions, HorizontalAlign, Padding, Point, Position, Ver
 use std::any::Any;
 use theme::Theme;
 use widget::{self, Widget};
+use ::std::io::Write;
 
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Hey, thanks :)

I think using `unwrap` in this case is fine, as there really shouldn't be a case for error, and I believe `unwrap` prints the error for `Result`s anyway - any chance you can revert it back before merging?

Esption:
@mitchmindtree The only way it should fail is if it's unable to print to stderr for some reason (which is probably a sign of other problems), and I wasn't sure if an unwrap() was best, as I think it will cause the thread to panic.  Since it was doing println!() before and not panic!(), I figured not panic'ing was the desired behvaiour.

mitchmindtree:
Alright, no problems, thanks again :+1:

Issue #205, opened 2014-12-06
Add "disabled" state for widgets that accept input


cybergeek94:
Proposition: `.disabled()` modifier or an `.enabled(bool)` setter that makes the widget ignore all input when disabled. Additional modifiers for making the disabled state clear to the user (e.g. greyed out and permanently-depressed state of disabled buttons) might be a good idea, though that can be set manually by the caller.

Chiefly useful for:

* Button
    * When the action that the button is supposed to do is not available, e.g. disabling "Previous Page" button for a paginated element when on page one.
* Slider
    * Can double as a progress-bar if the user can't modify it. Works currently if the slider is updated quickly enough but it really should just be able to reject user input.
* TextBox
    * Sometimes better than a Label: looks better in some contexts, can also have UI elements positioned relative to it
* Toggle
    * When toggling the element would leave the program in an invalid state.

The ability to disable elements that accept input is ubiquitous amongst UI libraries. On the upside, the most basic functionality seems trivial to implement. I might issue a PR for this shortly myself.

mitchmindtree:
This is addressed by #429.

PR #429, opened 2015-05-08
Implemented `enabled` method for enabling and disabling user input from Widgets.

Esption:
#205 @cybergeek94 @mitchmindtree 

Implemented for all types that impl Widget (Except calling ```enabled()``` on ```Label`` which will just get ignored.)

Default state is ```true``` for all so it shouldn't break any existing code.

Tested all with examples/all_widgets.rs and it seems to work just fine.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,6 @@ pub use widget::slider::Slider;
 pub use widget::text_box::TextBox;
 pub use widget::toggle::Toggle;
 pub use widget::xy_pad::XYPad;
-pub use widget::Toggleable;
 
 pub use background::Background;
 pub use canvas::{Canvas, CanvasId};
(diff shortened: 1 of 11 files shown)


mitchmindtree:
Thanks for this :+1:

Could you please move the `enabled` method out of the `Widget` trait and add it directly to each widget's implementation? i.e.
```Rust
impl Button {
    ...
    
    /// Enable or disable the Widget from responding to user input.
    pub fn enabled(...)  {
        ...
    }
}
```
This will also allow you to remove the method from the `Label` widget.

The aim for the `Widget` trait is that it should only implement methods that are necessary for *all* widgets. Methods that are not needed by all widgets should either be implemented directly on the widget (as above) or in a unique trait (normally a unique trait is only preferable if the trait would also provide useful defaults - see [`Positionable`](https://github.com/PistonDevelopers/conrod/blob/master/src/position.rs#L98) for an example of this).

Esption:
@mitchmindtree Done!

mitchmindtree:
Perfect, thanks a lot! You're on a roll :smile:

Issue #430, opened 2015-05-09
Change `UiId` to `WidgetId` to distinguish more clearly from `CanvasId`.


mitchmindtree:
Seeing as both widgets and canvasses have their own caches, perhaps it would be better to more clearly distinguish the `UiId` (which is an index into the `widget_cache`) from the `CanvasId` (an index into the `canvas_cache`) and call it `WidgetId` instead.

mitchmindtree:
We'll also need an enum to distinguish between the IDs for checking user input capturing (i.e. a Canvas may capture mouse input when being dragged, or key input for shortcuts). Perhaps we can use UiId for this:
```Rust
pub enum UiId {
    Widget(WidgetId),
    Canvas(CanvasId),
}
```

themooserooster:
I'd like to give this one a crack.

mitchmindtree:
@themooserooster I've already started on this - it's turning out to involve a lot of other small changes with the capturing of different Ui elements etc, I'll probably just include it in my next PR along with the Floating canvas. Thanks for the intent though!

themooserooster:
Ok, fair enough. I'm just trying to get started with Rust, so this seemed like a project worth contributing to.

mitchmindtree:
Closed by #433

PR #433, opened 2015-05-14
Implemented Floating Canvas type, made `Widget::{update, draw}` pure functions and more.

### LINKED ISSUE #424: Make `Widget` methods pure functions for clarity.
Currently, the `&mut Ui` and `UiId` is required to be available within the `update` and `draw` methods for a few annoying reasons:
1. determine the xy position of the widget (considers other widgets and canvasses for relative positioning)
2. retrieve any input (i.e. mouse, keyboard)
3. gain access to the `Theme`
4. determine the width of some text (using the character cache)

Solutions:

1. Could be calculated prior to calling the `update` if we had access to positioning, alignment and size of the Widget. This could be done by adding `get_position -> Position` and `get_align -> (VAlign, HAlign)` methods to the `Positionable` trait and then requiring that for a type to implement `Widget`, it must also implement `Positionable` and `Sizeable` (both of which are implemented on all widgets anyway).

2. Rather than passing in the `Ui` directly, we could pass in a `UserInput` type that stores references to the `Mouse` state as well as the recently entered key/text buffers.

3. The `Theme` could just be passed in by ref

4. This is the annoying one. Ultimately we would pass in `char_cache: &C where C: CharacterCache`. The reason we need the `character_cache` as mutable is because if we want the dimensions of a character that isn't yet cached, the cache needs to construct and store the glyph for that character before it can return them. There are a couple of ways we could solve this:
    - Pre-load all possible glyphs for every font and size upon construction of the CharacterCache. This would allow us to call the `.character` and `.width` methods by ref rather than mut_ref, as we know we wouldn't have to construct any more glyphs. The tradeoff is that we would probably have to limit the sizes that can be used within Conrod, as storing every glyph for every possible size would likely be ridiculously memory intensive and would lag startup.
    - Place the `Ui`s `character_cache` inside an `Rc<RefCell<>>`. This would still require that the character_cache is passed in mutably, however it would still allow us to borrow other parts of the `Ui` (i.e. for the hypothetical `UserInput` type) without causing any ownership issues.
    - Change the `CharacterCache` `.character` and `.width` methods to take `&self` instead of `&mut self`, and force implementations to use a lock or `Rc<RefCell<<>>` over the underlying `HashMap` itself. This is my personal favourite, as it would likely improve the `CharacterCache`'s usability in many more places than just Conrod.

All of these changes combined could change `update` from this:
```Rust
fn update<C>(self,
             prev: &State<Self::State>,
             current_style: &Self::Style,
             ui_id: UiId,
             ui: &mut Ui<C>) -> State<Option<Self::State>>
    where C: CharacterCache;
```
to this:
```Rust
fn update<C>(self,
             xy: Point,
             dim: Dimensions,
             prev: &State<Self::State>,
             current_style: &Self::Style,
             theme: &Theme,
             input: &UserInput,
             char_cache: &C) -> Option<Self::State>
    where C: CharacterCache;
```

and `draw` would change from this:
```Rust
fn draw<C>(new_state: &State<Self::State>,
           current_style: &Self::Style,   
           ui: &mut Ui<C>) -> Element     
    where C: CharacterCache;              
```
to this:
```Rust
fn draw<C>(new_state: &State<Self::State>,
           current_style: &Self::Style,
           theme: &Theme,   
           char_cache: &C) -> Element     
    where C: CharacterCache;              
```

mitchmindtree:
The main change here is that we now have a draggable canvas type! See the updated canvas.rs example for a demo. It's still missing a few features (scroll-ability, resizing, position bounds) but it's a nice start :)

Also, #424 has been implemented now so the `Widget` trait's update and draw signatures have changed.

See the commit messages for any more details.

Closes #424, #380 and #430

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: rust
 
 rust:
     - nightly
+    - beta
 
 notifications:
     irc: "irc.mozilla.org#piston-internals"
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -24,6 +24,7 @@ use conrod::{
     Colorable,
     DropDownList,
     EnvelopeEditor,
+    Floating,
     Frameable,
     Label,
     Labelable,
@@ -37,6 +38,7 @@ use conrod::{
     Toggle,
     Ui,
     UiId,
+    WidgetId,
     Widget,
     WidgetMatrix,
     XYPad,
@@ -163,6 +165,9 @@ fn draw_ui<'a>(gl: &mut GlGraphics, ui: &mut Ui<GlyphCache<'a>>, demo: &mut Demo
     // Draw the background.
     Background::new().color(demo.bg_color).draw(ui, gl);
 
+    // Construct a floating canvas in the centre of the screen.
+    Floating::new().label("Testing...").label_color(white()).set(0, ui);
+
     // Calculate x and y coords for title (temporary until `Canvas`es are implemented, see #380).
     let title_x = demo.title_pad - (ui.win_w / 2.0) + 185.0;
     let title_y = (ui.win_h / 2.0) - 50.0;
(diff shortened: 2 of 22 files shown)


mitchmindtree:
Woohoooo, builds successfully on beta!

mitchmindtree:
Here's a pic of two small `Floating` canvasses (from the canvas.rs example) overlapping.

screen shot 2015-05-14 at 11 55 09 am

mitchmindtree:
Going to merge this for now as travis is passing on both beta and nightly, lemme know if anyone has any issues :)

viperscape:
:+1:

bvssvni:
:+1:

PR #440, opened 2015-05-18
Implement Scrolling for DropDownList (WIP)

mitchmindtree:
**WIP - Do not merge this yet.**

The functionality is implemented and the scrollbar is usable, however the view of the `DropDownList`'s `Element` is not yet being cropped (i.e. you can see the items scrolling above and below what the visible area should be).

To fix this I think we need to add a `crop` method to `Element` which returns another cropped version of the `Element`. The relevant issue for the elmesque crate is [here](https://github.com/mitchmindtree/elmesque/issues/17).

@bvssvni am I correct in thinking I should be able to use `Viewport` to crop the view? Or should I be looking at something else?

Edit: ahhh I think what I'm actually after is `draw_state` - I see the `scissor` option now which looks like what I'm after :)


mitchmindtree:
Ok, I'm going to merge this now as it has some fixes for the latest elmesque changes. The cropping is still not working properly for some reason, but should be fine to fix in a separate PR.

PR #448, opened 2015-05-20
Moved Dimensions into Style structs for easier customisation using Themes

mitchmindtree:


mitchmindtree:
Travis is super slow tonight, all of these changes are basic and shouldn't cause any breakages, going to merge.

Issue #40, opened 2014-08-11
Auto-assign UIID to `<widget_name>::draw` functions.


mitchmindtree:
At the moment, a unique UI_ID (User Interface ID) must be passed into each unique `<widget_name>::draw` call. This UI_ID is used by the `UIContext` to store the state for the associated widget and retrieve that state each time the draw function is called.

Currently, it is up to the user to make sure that they pass in a unique UI_ID for each widget that they draw (see the "all_widgets" example). This can be annoying for the user to keep track of, and makes it difficult to build parent widgets out of several child widgets, as each child may also require their own UI_ID.

An ideal solution would be to somehow automatically statically assign a unique UI_ID for every `<widget_name>::draw` call. Perhaps this is possible with a macro of some sort? Maybe one that generates a unique number every time it is called within code (I don't know if this is possible)?

This isn't an issue for retained-mode UI as the state is stored within unique widget structs meaning a `UIContext` is unnecessary. However, I'd much prefer we managed to solve the UI_ID issue, as in most cases retaining widget structs for every widget's state seems to be overkill, especially when almost all of their state can be procedurally derived from the data and context they are associated with anyway.

bvssvni:
In 90% of cases you want to display a list of controls, and the same goes for interfaces on mobile platforms. I think unique numbers is fine for this.

bvssvni:
Closing.

viperscape:
Seeing how there is a desire for a [widgets-id macro](https://github.com/PistonDevelopers/conrod/issues/451), I think it would be beneficial if the uiids were generated and given as tokens for each widget. It can be complex to properly associate widget ids with their corresponding widget manually, especially if widgets are [dynamically created](https://github.com/viperscape/mush/blob/graph_backend/src/toolpane.rs#L56). Is it possible to store widget ids in the ui context, perhaps as a hashset?
I am considering a work-around but I'm not sure if it's possible to destroy a widget-id and mark for reuse, otherwise continually adding widget-ids during runtime will ultimately fail-- I assume; it's also overhead for the user, just as 'thinking about a coordinate system' would be.

Something like: 
``` rust
HashSet<(usize,u32)> //widget id, random::<u32>, though u64 might be better
```
[...]

Issue #451, opened 2015-05-21
Macro for generating Widget and Canvas Ids


mitchmindtree:
I imagine invocation would look something like this:
```Rust
widget_ids! {
    FOO,
    BAR,
    MATRIX <- 64,
    BAZ,
}
```

Which would expand into
```Rust
const FOO: WidgetId = 0;
const BAR: WidgetId = 1;
const MATRIX: WidgetId = 2;
const BAZ: WidgetId = 66;
```

This would save the user from doing something like:
```Rust
const FOO: WidgetId = 0;
const BAR: WidgetId = FOO + 1;
const MATRIX: WidgetId = BAR + 1;
const BAZ: WidgetId = MATRIX + 64;
```

I believe this should be possible with a [recursive macro](https://doc.rust-lang.org/book/macros.html#recursive-macros) to achieve the module level counting behaviour. I've had a go at this but kept running into the following compile error:
```
error: `$current_id:ident` is followed by a sequence repetition, which is not allowed for `ident` fragments
```

Would love it if someone would like to have a go at this :+1:

bvssvni:
Some ideas:

- Combine `&'static str` and usize
- `.begin` and `.end` methods for nested hierarchies
- Instead of incrementing, use `0` for `+1`

bvssvni:
What about storing nested `Ui`?

jarrett:
[This gist](https://gist.github.com/jarrett/e330ae21b40e85753a5c) is very close to what @mitchmindtree proposed, with an alternative to the proposed `MATRIX <- 64` notation.

I'm not sure if the `MATRIX <- 64` notation will be possible with Rust's macros. I think we need repetition in the matcher. Apparently, Rust's [macro grammar](https://doc.rust-lang.org/grammar.html#macros) doesn't allow variants inside a repetition matcher. That is, it doesn't seem to support a matcher like this:

    $($tail:ident | $tail -> $offset:expr)

...which I think we might need for `MATRIX <- 64`. I could be wrong, though.

bvssvni:
@jarrett :+1: You can use `start: X` multiple times, right?

jarrett:
@bvssvni you can call `widget_ids` multiple times, passing in different `start` values. But I don't think you can use `start` more than once in a given call to `widget_ids`. It's the same problem as with `MATRIX <- 64`: It would require matching on multiple possibilities inside the repetition.

mitchmindtree:
[...]

PR #454, opened 2015-05-22
[wip] auto uiid for ui

viperscape:
This is a first attempt at a possible [auto uiid](https://github.com/PistonDevelopers/conrod/issues/40) for ui contexts. It's slightly heavy handed when it comes to use, see [tests](https://github.com/viperscape/conrod/blob/auto-uiid/tests/lib.rs#L72) for example. It would be better if widgets had a trait method like ```set``` called ```ifset``` that determines if ui contains said widget. ~~I didn't want to make any breaking changes, so I never implemented that.~~

Additionally, this mixes with manual widget ids to an extent-- but could be complete if instead of [only tracking](https://github.com/viperscape/conrod/blob/auto-uiid/src/ui.rs#L111) widget ids we checked widget_cache first for preexisting ids. That's more complicated because canvas and widget cache is separate.


viperscape:
ifset method for widget looks like this:
``` rust
        Button::new().dimensions(80.0, 80.0).label("b4").react(|| { }).ifset(&b4, &mut ui);
```
where b4 is a button previously added to the ui:
``` rust 
    let b4 = ui.add_uiid();
```
[see tests](https://github.com/viperscape/conrod/blob/auto-uiid/tests/lib.rs#L63)

viperscape:
I don't know why travis won't build, says XNotSupported..

```

     Running `/home/travis/build/PistonDevelopers/conrod/target/debug/lib-58355e208c2d9444`
running 2 tests
test test_auto_uiid ... FAILED
test test_basic ... FAILED
failures:
---- test_auto_uiid stdout ----
	thread 'test_auto_uiid' panicked at 'called `Result::unwrap()` on an `Err` value: XNotSupported', /home/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libcore/result.rs:729
---- test_basic stdout ----
	thread 'test_basic' panicked at 'called `Result::unwrap()` on an `Err` value: XNotSupported', /home/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libcore/result.rs:729
failures:
    test_auto_uiid
    test_basic
test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured
thread '<main>' panicked at 'Some tests failed', /home/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-linux/build/src/libtest/lib.rs:259
```

mitchmindtree:
Sorry chap, I'm currently out of town atm but will be back later tonight to have a look

viperscape:
@mitchmindtree no rush. I am also interested to know if canvas cache and widget contain unique ids between the two, as I don't understand why they are separate. Thanks, enjoy the weekend!

viperscape:
I think the crate 'window' was removed from the deps in conrod, because I cannot build this anymore. I originally based this off of [conrod's benches](https://github.com/PistonDevelopers/conrod/blob/master/benches/draw.rs), so benches might not be building either. This PR didn't seem to get any traction, and I don't have much time to invest right now so I'm going to close the PR for now, feel free to reopen if you think this deserves more attention.

bvssvni:
@viperscape This deserves more attention!

bvssvni:
We changed to crates.io dependencies some days ago which might be the reason it does not compile.

viperscape:
oh okay, noted! I'll try and work on this some more then. I think formal tests should be available for conrod, I didn't see any so I started a tests folder. I think the previous Travis failure is because the tests require a head, I don't know how to do a headless test for conrod. Hope that makes sense.

bvssvni:
- `ui.get_uiid()` could be renamed to `ui.get_id()` or just `ui.id()`
- `.ifset` could be `.if_set`

bvssvni:
Do I understand this correctly: When you are writing a widget library, you can create a sub set of ids that are guaranteed to not collide with other ids? Is this what happens when you swap the ui id handler?

viperscape:
> ui.get_uiid() could be renamed to ui.get_id() or just ui.id()
.ifset could be .if_set

That's fine, I kept uiid to try and match the library, I prefer id as well. I was thinking ```ifset``` could be ```setif``` to more closely match ```set```, perhaps ```set_if``` then?

> When you are writing a widget library, you can create a sub set of ids that are guaranteed to not collide with other ids? Is this what happens when you swap the ui id handler?

Well not a library, perhaps that's possible too, but as a user the widgets can be registered without worrying about ids. The idea is to register your widget ids with the ui, then reference them during the widget creation ```ifset```. 
If there is a desire to mix manual widget ids and auto, mostly all that needs to happen is for a check to be done during uiid-handlers registration to see if the id exists in cache already, I plan on adding that but I am unsure of the differences between widget_cache and canvas_cache. For instance, can a widget id of 0 exist while a canvas id of 0 also exists in cache? I should check that. To make this mix of auto/manual more reliable, a reservation-range could be made in conrod. So when a user starts their manual widget ids at 0, let's say, the id actually starts (internally) at perhaps 1000, or higher. The range 0-range_max ids, internally in conrod, could be made for registered ids to prevent any collisions between manual and auto-id widget ids, though I don't know if any of that is truly necessary but is a good safe guard.
As far as a widget library that has sub widgets, I think as long as the ui context was passed during construction, then the widget library could also register its sub widgets and use the registered-id ```rid``` references during the render loop.

The steps of use are:
- [register id](https://github.com/viperscape/conrod/blob/auto-uiid/tests/lib.rs#L66)
- [set widget with special set_if](https://github.com/viperscape/conrod/blob/auto-uiid/tests/lib.rs#L83)

I want to get tests working though, as we should have a way to run conrod tests-- I'm not even 100% sure that the panics (which route/print to console) in conrod would fail a test. I should add a [should_panic] test too then.

One of the things to think about is how to get widget id creation to be available during the render loop, so some widgets could be registered, used, and removed all within the loop-- dynamically so to speak.

bvssvni:
It seems to be two kinds of using ids:

1. When you iterate through a list and create a widget for each item. In this case you want to just increment the id.
2. When you reserve an id. You get the same state no matter which order widgets are rendered. In this case it does not matter what the id is, as long as it is unique.

I think the problem we need to solve is how to cover both these use cases, and what properties we want when widgets relate to each other across these two use cases.

bvssvni:
The semantics between those two use cases are different:

1. When you increment an id, you get a unique number, but it is also deterministically the same number if the root starts with the same id the second time you repeat the operation.
2. When you reserve an id, you get a unique number, but it is different each time you reserve a new number.

Maybe we can replace the increment with a hash function, such that you can reserve using a random number and then "increment" it to get the deterministic behavior.

bvssvni:
If we use an `u64` then we could split it into chunks of 10 million for incremental use and still have 1844674407370 to pick from. If we pick a random number among those, with an optional seed, then we can use the id of a parent as the seed to get deterministic reservation for the children of the parent.

bvssvni:
If we have 100 000 parents then the probability of collision with a perfect random generator is:

```
x=100000
1-(1-1/1844674407370)^x = 5.42e-8
```

If we use 1 million chunks for incremental use then the probability drops with a magnitude to 5.42e-9.

I am sure we can adjust these factors to get the desired properties.

bvssvni:
I'll move the analysis to another issue.

bvssvni:
I deleted my comments and opened https://github.com/PistonDevelopers/conrod/issues/467

mitchmindtree:
Hey thanks for going ahead with discussion etc! This PR slipped my mind, apologies!

I'm still not 100% clear on the extra benefits that this method of ID generation will provide? Could you perhaps provide an example in which this dynamic method would work, in which the [`widget_ids!` macro](https://github.com/PistonDevelopers/conrod/issues/451) would not (or be much more painful)?

One of the more complex GUIs I can think of would be a dynamic graph of widgets - here are a couple solutions that come to mind in terms of handling this within conrod:

1. Treat the whole `graph` as a single widget, i.e. design a `GraphWidget` that can take a `&mut Graph` and visualise/control it. This would then only take a single `WidgetId` no matter how large the graph.

2. If for some reason this approach can't be taken, normally a user would have a safe estimate of the max number of nodes that may exist at once. With `widget_ids!` I imagine you could do something like:
```Rust
const MAX_NUM_OF_NODES: WidgetId = 512;

widget_ids! {
    FOO,
    GRAPH_TITLE,
    GRAPH_NODE <- MAX_NUM_OF_NODES,
    BAR,
}
```

If the `widget_ids` macro were to go ahead, the `FOO <- SOME_USIZE` syntax would indicate the number of widget IDs reserved for some widget `FOO`. In our above example, this means we could get the unique ID for each node just by adding `GRAPH_NODE` with the node's [unique NodeIndex](http://bluss.github.io/petulant-avenger-graphlibrary/target/doc/petgraph/graph/struct.NodeIndex.html#method.index) and never have to worry about a conflict.

In the case that we anticipate our graph's nodes will be constantly changing, perhaps we could add a
```Rust
ui.reset_widget_id(ID);
```
method which would set the cached position back to an empty state, allowing the ID to be used for whatever the next widget that is set via that ID is. We could also have a version that took a range of IDs. i.e. if we wanted to clear all our node widgets:
```Rust
ui.reset_widget_ids(GRAPH_NODE..GRAPH_NODE + MAX_NUM_OF_NODES);
```

Option 1 would be the ideal, but I don't think option 2 is so bad in the case that for some reason option 1 isn't doable.

I'm certainly open to the idea of making widget id generation more dynamic! I guess I'm just struggling to imagine a problem that it can solve that can't be handled fairly easily with a list of `const` IDs.

Sorry again for taking so long to drop by!

viperscape:
These are good ideas. You're right, the use case is probably limited-- and since [mush](https://github.com/PistonDevelopers/mush) was using conrod I came in to this need. This is just one way to solve the problem, I think a range of IDs works better, and the random-seed for id generation is a neat idea too. Aside from graph libraries, I'm not sure how useful dynamic widget generation would be, perhaps basic 2d-games? Thanks for the feedback!

bvssvni:
@viperscape Thanks for the input! It seems we will stick with the current design for now.

Closing.

PR #477, opened 2015-06-12
Fix DropDownList issue where some items wouldn't display when scrolled to the bottom of a long list.

mitchmindtree:


bvssvni:
Nice!

PR #478, opened 2015-06-14
Fix bug where slider would glitch on previous value if mouse is not over canvas.

mitchmindtree:


mitchmindtree:
Travis failed on nightly building the docs for the clock_ticks crate, but built fine on stable - going to merge.

PR #508, opened 2015-07-14
Graph implementation and consolidation of `Widget` and `Canvas`

### LINKED ISSUE #494: Allow custom `Widget`s to be composed of other `Widget`s.
At the moment, I'm quite sure that this is not possible without the parent `Widget` reserving multiple `UiId`s for each of its child `Widget`s. This is because Conrod's `UiId` system is linear, and every `Widget` requires its own `UiId`.

If `Widget`s are to be composable from other `Widget`s, then this would imply a more tree-like system.

Perhaps each `Cached` widget could also have its own cache of widgets that it manages, rather than having the `Ui` manage *all* widgets?

This would also mean we could remove the `canvas_cache`, and consolidate the different `Canvas` kinds as `Widget`s and store everything in the `widget_cache`.

It would also make designing custom widgets *much* easier, as users would be able to use pre-existing widgets with ease.

This would be a large refactor, but definitely worth it... Going to put some thought into this.

It would be nice to have this done prior to doing the custom widgets example #472.

### LINKED ISSUE #499: Use a `petgraph::Graph` instead of a `Vec` for the `Ui`'s internal `widget_cache`
Related to #494 

At the moment there is a single `Vec<widget::Cached>` used as the `widget_cache` within the `Ui`. I propose that we instead use [petgraph](https://github.com/bluss/petulant-avenger-graphlibrary)'s [`Graph` type](http://bluss.github.io/petulant-avenger-graphlibrary/doc/petgraph/graph/struct.Graph.html) for the following reasons:
- We could represent the UI tree using Directed `Edge`s (i.e. for `a -> b`, `a` is the parent of `b`).
- We could easily check whether or not setting a widget's parent would create a cycle.
- Look-up will still be O(1) as `Graph` uses `Vec` internally.
- Adds flexibility for describing relationships between different widgets (using an `Edge` enum with multiple variants).
- Safely handle ownership and multiple-node access by using provided Graph methods for walking edges, etc.

One issue is that `Graph` produces its own set of indices. We could solve this by creating an `IndexMap` type, which maps a `WidgetId` to its associated `NodeIndex` and vice versa. This could be done with a `Vec<NodeIndex>` and a `Vec<WidgetId>`.

We could also use the `IndexMap` to distinguish between internal widgets (widgets used by a widget designer to compose some other widget) and public widgets (widgets `.set` into the `Ui` via the API with some `UiId`). This could be done with two `Vec<NodeIndex>` and a `Vec<Either<InternalId, PublicId>>`.

### LINKED ISSUE #500: Add a `widget::CommonBuilder` struct to hold data needed by all Widget types.
This would vastly simplify the process of making custom widgets.

Rather than requiring Widget types to implement `Positionable` and `Sizeable` (traits common to all `Widget`s), we can instead store positioning, alignment, depth and sizing data in a `CommonBuilder` struct and require that types implementing `Widget` implement a method that returns a `&mut CommonBuilder`. We could then do blanket impls of `Positionable` and `Sizeable` for all `T: Widget`.

### LINKED ISSUE #507: Should the `Widget::update/draw` args be combined into single `Update/DrawArgs` types?
At the moment the args for the `Widget::update` and `Widget::draw` methods are pretty verbose and it can be a bit of a pain when writing custom `Widget` implementations. Perhaps we can do something like the following:

```Rust
    /// Update the state of the Canvas.
    fn update<'a, C>(mut self,
                     prev_state: &widget::State<State>,
                     xy: Point,
                     dim: Dimensions,
                     input: UserInput<'a>,
                     style: &Style,
                     theme: &Theme,
                     glyph_cache: &GlyphCache<C>) -> Option<State>
        where C: CharacterCache;
```

would become

```Rust
    /// Update the state of the Canvas.
    fn update<'a, 'b, C>(mut self, args: UpdateArgs<'a, 'b, C>) -> Option<State>
        where C: CharacterCache;
```

(same goes for `Widget::draw`)

This should also reduce the number of imports needed.

mitchmindtree:
~~**DON'T MERGE** (not done yet)~~

Apologies for the epic PR! The `Graph` implementation leaked into a lot of other issues so it made sense to address a few things at once.

Graph implementation is mostly done, still need to add a way to distinguish between floating widgets in the graph.  Closes #499 .

Consolidated the `Widget::update` and `Widget::draw` args into their own structs, which has tidied up the widget implementations quite a bit, and results in a lot less necessary imports. Closes #507 .

Added a `CommonBuilder` type which stores all builder data that is common to all widgets (i.e. positioning, size, render depth, parent id) which allows us to blanket impl the `Positionable` and `Sizeable` traits and get rid of the impls in each of the individual widget modules :) Closes #500 .

I'm about to get started on the interface for setting internal widgets using the `Widget` trait, so that a widget may be composed of other widgets without occupying the `WidgetId` space. Once this is done I'll re-implement the `Split` and `Floating` canvas variants on top of the newly added `Canvas` `Widget`. Closes #494 .


mitchmindtree:
Ok, should be done!

I think everything's working nicely - all the examples behave just as they did before.

Checkout the diffs for the examples to see the tweaks that have been made to the user API - main points are:
- There's no more distinction between `CanvasId` and `WidgetId` (therefor there is no longer a `UiId`, and only a `WidgetId`). Canvasses are now implemented entirely as Widgets!
- There is no more `Floating` canvas type - `Canvas` now fills this role and offers a `floating` method for producing the original `Floating` behaviour (see canvas.rs example).
- The `Split` canvas type is now just a wrapper around the `Canvas` widget (is still used the same as it was before though).

Might need to expose some of the new types that have appeared in the widget module for folks designing custom widgets, but I think it might be best to review and do this in a separate PR :+1:

There's still no method for adding widgets internally (without occupying ID space), but we're a huge step closer now that we have the graph implemented and the consolidation of widget/canvas!

bvssvni:
Looks good! Merging.

bvssvni:
:tada:

src/graph/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+        };
+
+        // Walk our incoming edges and remove any that aren't connected to our requested parent.
+        let mut incoming_edges = graph.walk_edges_directed(node_idx, pg::Incoming);
+        let mut already_connected = false;
+        while let Some((in_edge_idx, in_node_idx)) = incoming_edges.next_neighbor(graph) {
+            if in_node_idx == parent_node_idx {
+                already_connected = true;
+            } else {
+                graph.remove_edge(in_edge_idx);
```
bluss: Hi! (Hi mindtree!)

Looking at this crate since it's a new reverse dependency.

I wouldn't recommend removing edges while iterating edges. Petgraph's Graph is a safe datastructure, so this will never violate memory safety, but the next index in the iteration might point to something completely different (or be out of bounds, and iteration stops).

A section of petgraph's documentation is in boldface:

> NodeIndex and EdgeIndex are types that act as references to nodes and edges, but these are only stable across certain operations. **Removing nodes or edges may shift other indices.** Adding to the graph keeps all indices stable, but removing a node will force the last node to shift its index to take its place. Similarly, removing an edge shifts the index of the last edge.

It's a very bare bones datastructure with the philosophy that every graph user will need something different. We use indices instead of references to allow very powerful things, for example mixing traversal with mutating edge or node weights, possibly two at a time (index_twice_mut)

viperscape:
Not sure if this will help, but [mush](https://github.com/PistonDevelopers/mush) might also be something to look at, if just for ideas. It makes use of hashmaps and hashsets (for edges). It likely won't compile right now since all of these conrod changes have taken place. Great work, glad to see conrod is taking on a graph layout, many benefits I think

src/graph/mod.rs
mitchmindtree: @bluss to the rescue once again!

As it turned out, there can only ever be one edge at a time (as that method is the only publicly available way of adding edges) so it can't cause any errors as it stands, however it's certainly not clear and could easily cause issues in the future when refactoring etc!

I might just change this from `while let` to `if let` for now and add some decent docs.

Thanks for being lovely and looking out for your reverse deps!

bluss: Aha, that explains it well

PR #523, opened 2015-07-28
Fix error with dropdownmenu where it would panic if scrolled to bottom with no frame.

### LINKED ISSUE #520: DropDownList panics if is scrolled to bottom when there is no frame (see all_widgets example).

mitchmindtree:
Closes #520


mitchmindtree:
Merging as stable build was successful (only the nightly failed as it didn't start the build properly for some reason).

PR #524, opened 2015-07-28
Fix issue where cursor idx would be outside of text range if text cleared during react function.

### LINKED ISSUE #517: When clearing the text in a text box, it panics
Seems to be caused by https://github.com/PistonDevelopers/conrod/blob/master/src/widget/text_box.rs#L172

I haven't tested this myself (was reported on IRC), but worth checking the logic.

mitchmindtree:
Closes #517

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.7.0"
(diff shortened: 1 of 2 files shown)


stratact:
All good here! *whistles* :notes:

Issue #472, opened 2015-06-10
Create custom Widget examples.


mitchmindtree:
I think the following would really help users design their own `Widget`s:
1. A simple as possible example of making and using a custom `Widget`.
2. An advanced example of using multiple pre-existing `Widget`s to design a single new `Widget`. This is probably difficult at the moment, but we should work to make this as easy as possible.

m-bra:
This would help me too since I'm currently struggling creating a custom widget...
Using the internal widgets as reference does not work since they use other than just the `conrod` crate, for example to use `Scalar` one needs the `graphics` crate and I have no idea where to get that from

jarrett:
I'm working on a new file in the `examples` folder that will illustrate custom widgets.

jarrett:
I've added a custom widgets example on a [new branch](https://github.com/jarrett/conrod/tree/custom-widgets).

[Here's the example.](https://github.com/jarrett/conrod/blob/custom-widgets/examples/custom_widget.rs).

I had to [add some re-exports](https://github.com/jarrett/conrod/commit/313e618bf99d090ecf1dd6c9f23c3606f2233176#diff-b4aea3e418ccdb71239b96952d9cddb6) to the root module. That's because defining a widget requires some names that were previously private.

mitchmindtree:
Closed by #525.

PR #525, opened 2015-07-30
Re-exports to support custom widgets. Custom widget example.

jarrett:
See https://github.com/PistonDevelopers/conrod/issues/472#issuecomment-126503484.


mitchmindtree:
Awesome @jarrett ! Thanks so much for this :smile_cat:

Issue #439, opened 2015-05-17
Add an elmesque example


dcampbell24:
[Elmesque](https://github.com/mitchmindtree/elmesque) is currently the suggested high level 2d graphics crate to use with conrod, so an example showing how to use them together would be great to have in this repository.

cc #438

mitchmindtree:
Sure, I'll try to get around to this some time soon, but would love it if anyone else is willing to give this a go (I've got lots on my plate at the moment).

jarrett:
I'm about to attempt this on a real project. Related to https://github.com/PistonDevelopers/conrod/issues/519, I'll offer my example if I can get it working.

One question though: Conrod [constructs its own Elmesque renderer](https://github.com/PistonDevelopers/conrod/blob/386eb12564a765a41af124ca50fa385d615757cf/src/ui.rs#L358). For my custom Elmesque drawing outside Conrod, should I just construct another renderer? Will the two renderers interfere with each other? My guess it is should be fine, but I wanted to see if you have a recommendation.

jarrett:
I've written an example showing how to combine Conrod with custom Elmesque drawing:

https://github.com/jarrett/conrod/blob/elmesque-example/examples/elmesque.rs
https://github.com/jarrett/conrod/commit/c1a81e3ee53f990b61c9b9a1af9788c8f4049011

mitchmindtree:
@jarrett there's no issue with having two `elmesque::Renderer`s (it's just a wrapper around a piston2d-graphics `Context`, `Backend` and `CharacterCache`), however I think we'll be able to make this nicer by implementing an `.element` method on the `Ui` that constructs and returns the whole GUI's `Element` (rather than drawing it on the spot as does the `.draw` method). This would allow a user to use their own renderer and to position the gui using elmesque layout, etc

Closing this via #526

PR #526, opened 2015-07-30
Added Elmesque example and note in readme.

jarrett:
See https://github.com/PistonDevelopers/conrod/issues/439.


mitchmindtree:
Great! Hopefully we can make working with elmesque a little nicer in the future, but this is good for now :+1:

Closes #439

PR #531, opened 2015-07-31
Added `Ui::draw_if_changed` (big speedup!), updated examples and readme. 

### LINKED ISSUE #530: Change `Ui::draw` to only re-draw if any of the Widgets have new `Element`s.
So far Conrod has pretty poor performance as it redraws its graphics every frame!

For folks doing purely GUI applications, this is clearly not desired, and they'd almost definitely prefer to just redraw when some widget actually requires it.

I'm guessing that this would reduce conrod's --release CPU usage from around 25% right down to barely noticeable, as the vast majority of time is currently spent within the elmesque/piston-graphics draw calls.

It's very easy to add (I'm doing it now) but I just thought I'd open a discussion on what the best naming for this might be?

So far my plan is to change `Ui::draw` to only re-draw if some `Widget`'s `draw` function is called (which only gets called if it returns a new state from `update`. When a widget is updated with a new state, it calls `ui.needs_redraw` automatically which will cause the `Ui` to redraw all the `Element`s when `Ui::draw` is called. We could do something similar for the upcoming `Ui::element` method (mentioned here #527 @jarrett ) which could return `Option<Element>` - it would return `Some(Element)` if a redraw was needed, otherwise returning `None`.

The `Ui::needs_redraw` method would be public, so if for some reason a user needed conrod to redraw everything they could call this before calling `Ui::draw`.

An alternative would be to leave `Ui::draw` to behave as it does currently, and add a separate method called `Ui::draw_if_needed` which behaves as described above. Come to think of it, this is probably a better option as it has less chance of breaking people's code and is more descriptive of the behaviour. We would change all examples to use this method as they are purely GUI (there is no other graphics stuff going on other than conrod's).

Any other ideas/thoughts welcome :+1:

mitchmindtree:
Using `draw_if_changed` (in --release mode), both `canvas` and `all_widgets` examples avg at about %6.5 on my machine now as opposed to the ~25% when using `.draw` (yaaay :tada:)

Also added some hefty docs to `.draw` and `.draw_if_changed` to clarify the use case for each.

I also changed the behaviour of `Background` so that rather than drawing at its call site, it just stores the color in the `Ui` to be used for clearing the screen when `Ui::draw` is called. This stops it from drawing over the top of the widgets when using `ui.draw_if_needed`.

Closes #530

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.7.2"
+version = "0.7.3"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Available Widgets
 -----------------
 
 - Button
+- Canvas (Can be positioned manually or by using Splits or Tabs)
 - Drop Down List
 - Envelope Editor
 - Label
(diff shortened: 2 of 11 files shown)


bvssvni:
Looks good! Might want to change the `SAFE_REDRAW_COUNT` into a setting on `Ui`.

mitchmindtree:
@bvssvni yeah good call, i'll add a method for setting it manually but leave `SAFE_REDRAW_COUNT` as the default :+1:

PR #538, opened 2015-08-03
Add `.element` method to `Ui` and `Graph` for returning the `Element` as an alternative to `.draw`

jarrett:


mitchmindtree:
Nice! I think we could also add `Ui::element_if_changed` that returns an `Option<Element>` (to match `draw_if_changed`) but we can leave this for another PR :+1:

PR #539, opened 2015-08-04
Update Dependencies

stratact:


mitchmindtree:
TY :smile_cat:

stratact:
We have happiness! :wink:

bvssvni:
Notice that when a dependency increases in the first non-zero number, the library should also increase in the first non-zero number. See https://github.com/PistonDevelopers/piston/issues/931

mitchmindtree:
@bvssvni Haha yep, @stratact  checked with me and I mistakenly said 0.7.6 was ok, but fixed it by changing it to 0.8.0 :+1:

bvssvni:
@mitchmindtree Good :smile:

PR #542, opened 2015-08-13
Opt-in generalised scrolling

mitchmindtree:
~~**DO NOT MERGE** *yet*~~

Added
```Rust
.scrolling(bool) // bi-directional scrolling
.vertical_scrolling(bool)
.horizontal_scrolling(bool)
```
methods to the `Widget` trait.

If a scrolling method is called with `true` on some parent widget and the child `Widget`s would occupy a greater area than the parent `Widget`'s "`kid_area`", the `kid_area` will automatically become scrollable and the children widgets will be cropped to the scrollable area.

Still need to:
- [x] fix bug where max_offset changes when scrolling down too quickly
- [x] crop scroll area properly
- [x] only scroll if mouse is over a widget that is either the scrollable widget or one of its children


mitchmindtree:
This PR also updates the canvas.rs example to demonstrate scrolling by making the bottom button matrix twice the size vertically and scrollable. It's hard to capture it in a pic, but here's the button matrix scrolled down about half way, with the scrollbar on the right.

scrollable area

I'm about to do another commit that will change the scroll drawing behaviour so that it only draws if the mouse is over where the scrollbar track would be. This should make it a little more subtle / modern behaving. Another nice behaviour would be to have it appear only while the "mousewheel" (or trackpad, etc) is being scrolled, but this can be for another PR.

mitchmindtree:
There are still some issues with cropping on retina display for some reason (an issue for elmesque) but otherwise should be fine to merge.

PR #559, opened 2015-09-08
[WIP] Add keyboard events to textbox

TheNeikos:
This should satisfy a few of the checkpoints in #71, 

However I am not sure that the rest of the checkpoints are user-friendly, for example Ctrl+Left in any application I know goes left by a word, that is until it finds a whitespace after a word.

### DIFF
--- a/src/widget/text_box.rs
+++ b/src/widget/text_box.rs
@@ -7,7 +7,7 @@ use graphics::math::Scalar;
 use label::FontSize;
 use mouse::Mouse;
 use num::Float;
-use input::keyboard::Key::{Backspace, Left, Right, Return, A, LCtrl, RCtrl};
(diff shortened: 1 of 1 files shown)


TheNeikos:
Interesting... the build fails because it can't find `apt-get` on Mac OSX :v Nothing I can do to fix that!

mitchmindtree:
@TheNeikos looks great! Thanks a lot :) Yeah I think we'll need to fix up the `.travis.yml` to handle both, although I don't think we need to test on multiple platforms, as I'm fairly sure we're using entirely portable code.. maybe it's worth opening an issue to discuss.

btw just thought I'd mention that in this comment editor, Ctrl+Left/Right moves to the start or end of the line for me and option+Left/Right moves one word (I'm on OSX).

TheNeikos:
Interesting, because on (Arch)Linux and Windows both do the Ctrl+Left/Right move one word.

PR #565, opened 2015-09-28
Added `Widget::set_internal` method to allow for `Widget`s to be composed of other `Widget`s

mitchmindtree:
Finally, it should now be possible to set `Widget`s within your own custom `Widget`'s `update` method without occupying the public `WidgetId` space! This means we should now be able to compose new `Widget`s from other `Widget`s.

To make this possible, I needed to add a few things:
- The `Widget::set_internal` method which we can use to set internal `Widget`s without occupying the `WidgetId` space.
- The `UiCell`, a wrapper around the `Ui` which is passed into the `Widget::update` method via the `UpdateArgs` and allows us to call `set_internal`.
- The `WidgetIndex` which allows us to dynamically handle both public and internal index types.

The change also allowed us to remove the ugly `Widget::set_children` method - a previous hack to allow for setting other public widgets in the `Tabs` widget. The changes to `Widget::set` and the addition of `Widget::set_internal` now render this unnecessary.

This is a **breaking change**, but only for custom widget designers and the necessary changes should be easy to implement (see the changed widget modules in this PR for an example of how to update).


mitchmindtree:
Going to merge and publish this now as travis is passing - if anyone finds any issues, feel free to post an issue :smile_cat:

bvssvni:
Wow, this was a great PR. :+1:

mitchmindtree:
@bvssvni thanks! Currently reworking `DropDownList` to use pre-existing widgets rather than making its own buttons and scrolling... it's about half the size now! Will hopefully have a PR in the next few hours.

PR #570, opened 2015-09-29
Massively simplify `DropDownList` by composing it from other `Widget`s. Tweak Widget API with lessons learned.

### LINKED ISSUE #566: Simplify DropDownList using latest Widget designer features.
DropDownList should be a great first widget for trying out the new `set_internal` method. Rather than drawing it's own rectangles for each button, and calculating it's own scrolling information, it should be able to be re-implemented as a Canvas widget with Button widgets on top which would use conrod's automated scrolling to achieve its scroll behaviour.

mitchmindtree:
`DropDownList` is now roughly half the size (its `update` method is much shorter, and it's `draw` method is now a one-liner) with all the same behaviour! I think this is a nice proof-of-concept and a good example of how easy it should be to compose new `Widget`s from other `Widget`s.

As expected, I had to make some tweaks to the widget module as I ran into different issues doing this for the first time:
- Turns out `set_internal` is totally unnecessary! By making `Widget::set` generic over any type that impls `Into<Index>`, we can use it for setting both `Public` (`WidgetId`) and `Internal` (`NodeIndex`) indices. This should keep the API much cleaner and consistent between widget designing and widget using.
- Sometimes a parent `Widget` needs to know whether or not the mouse was released, even though it may not be capturing it or hovered - thus I added a `global_mouse` field to the `UserInput` struct.
- In some occasions it can be annoying for a `Widget` to track whether or not a mouse was pressed or released, so I added `was_pressed` and `was_released` flags to the mouse's `ButtonState`.
- The std::error messages when two widgets simultaneously try to capture input is unnecessary and can be annoying even if the resulting behaviour is correct, so I removed it.

This is a slight **breaking change** due to the removal of the `set_internal` method (despite its brief life) and the change to the `mouse::ButtonState` struct.

Note: there is still an issue where the `DropDownList` doesn't crop properly to its scroll area - it now also eternally captures the mouse if you click one of the buttons outside the proper scroll area. The over-arching bug existed before this however, so it will be addressed in a separate PR.

EDIT: Also, I changed the last arg  of `DropDownList`'s `react` function to take a `&str` instead of a `String` - it was an unnecessary allocation seeing as a user can easily call `.to_owned()` if they require a `String`. Oddly however, this does now require that the user writes the type for the `&str` argument when calling `DropDownList::react` with a closure (otherwise the compiler seems to complain about the closure's lifetimes not matching that of the FnMut bound required by `DropDownList::react`'s arg type).

Closes #566

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.15.1"
+version = "0.16.0"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -355,7 +355,7 @@ fn draw_ui(c: Context, gl: &mut GlGraphics, ui: &mut Ui, demo: &mut DemoApp) {
         .frame_color(ddl_color.plain_contrast())
         .label("Colors")
         .label_color(ddl_color.plain_contrast())
-        .react(|selected_idx: &mut Option<usize>, new_idx, _string| {
+        .react(|selected_idx: &mut Option<usize>, new_idx, _string: &str| {
             *selected_idx = Some(new_idx)
         })
         .set(COLOR_SELECT, ui);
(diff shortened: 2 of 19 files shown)


mitchmindtree:
Travis is happy - going to merge :+1:

PR #577, opened 2015-10-02
Fix various scrolling issues. Add scrolling to the canvas in all_widgets. Changed WidgetMatrix. Added Range and Rect. Improved `set_widget` order. (See comments and commits for deets).

### LINKED ISSUE #578: Make `WidgetMatrix` an actual `Widget`
Rather than requiring the user to use public `WidgetId`s for each widget in the `WidgetMatrix`, have the `WidgetMatrix` pass its own unique `NodeIndex`s to the widget via the closure arguments.

This should also fix its scrolling behaviour.

mitchmindtree:
**WIP** (don't merge just yet)

- [x] Consider relative positioning properly in scroll_offset calculation.
- [x] Fix horizontal scrolling.
- [x] Make `WidgetMatrix` an actual `Widget` (which will fix its scroll behaviour). Closes #578.
    ~~- Turns out `impl`ing `Widget` for `WidgetMatrix` requires moving the `<C>` generic from the `update` method to the `Widget` trait itself. This is because `WidgetMatrix` holds some closure that has a `UiCell<C>` as an arg, and the C must match the same C given by the `UiCell<C>` passed to the update. I've been hoping to avoid this (mainly just because `Widget` trait looked so pretty without generic params) but it does make logical sense to make the change... so I might as well make the change along with this PR also.~~
- [x] In `set_widget`, update the widget's cached xy coords and dimensions before `Widget::update` is called, so that child widgets can correctly calculate scroll offset when set within `Widget::update`.
    - [x] Calculate `kid_area` prior to calling `Widget::update`
    - [x] Calculate the scrolling prior to calling `Widget::update`
    - [x] Split the `Graph::update_widget` method into `Graph::pre_update_cache` and `Graph::post_update_cache`. This will allow us to cache information about a widget necessary for the child widgets *prior* to calling `Widget::update` (in which child widgets are `set`).
~~- [ ] Move `CharacterCache` generic from `Widget::update` method to the `Widget` trait itself.~~
- [x] Fix issue where the scroll Bar's max_offset is slightly short when visible area is initially outside of the kids' bounding box.
- [x] Fix scrolling behaviour for floating widgets' children.
- [x] Fix bug causing canvas.rs initial scroll offset to be out of bounds.
- [x] Fix issue introduced in prev commit where Tabs kid area seems to be incorrect.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.19.0"
+version = "0.20.0"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
@@ -22,16 +22,16 @@ path = "./src/lib.rs"
 
 
 [dependencies]
-bitflags = "*"
+bitflags = "0.3.2"
 clock_ticks = "0.0.6"
 elmesque = "0.9.0"
 json_io = "0.1.2"
 petgraph = "0.1.10"
 pistoncore-input = "0.8.0"
 piston2d-graphics = "0.10.0"
-num = "*"
-rand = "*"
-rustc-serialize = "*"
+num = "0.1.27"
+rand = "0.3.11"
+rustc-serialize = "0.3.16"
 vecmath = "0.2.0"
 
 [dev-dependencies]
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -48,6 +48,7 @@ use piston::event_loop::{Events, EventLoop};
 use piston::input::{RenderEvent};
 use piston::window::{WindowSettings, Size};
 
+
 type Ui = conrod::Ui<GlyphCache<'static>>;
 
 /// This struct holds all of the variables used to demonstrate
@@ -128,6 +129,7 @@ impl DemoApp {
     }
 
 }
+
 
 fn main() {
     let opengl = OpenGL::V3_2;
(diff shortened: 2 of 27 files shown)


mitchmindtree:
Just thought I'd mention this is one of those PRs that leaks into a lot of areas - I'll be constantly editing the points to address in the original comment, and will be sure to rename the title if necessary before merging.

bvssvni:
I'm waiting with updating dependencies until this get merged.

mitchmindtree:
@bvssvni sorry I'm taking so long, getting this scrolling behaviour logic right is proving to be incredibly tricky. Feel free to update, I don't mind merging with your changes if I'm not done in time :+1:

mitchmindtree:
FINALLY!

Lots of improvements in this, namely:
- Scrolling should now work properly when the initial view rect is outside the kids bounds.
- Added scrolling to the main canvas in the all_widgets example, so if you resize the window smaller, the canvas becomes scrollable automatically.
- Add `Range` and `Rect` positioning types to make it easier to reason about widget placement, padding, etc, etc. Also updated widget `update` method to take a `rect` arg rather than `xy` and `dim` separately, however many still aren't utilising its functionality to the best potential (everything works though, so not a big deal).
- Improved the order in which the huge, imperative `set_widget` function does everything! The biggest breakthrough was splitting the cache update into "pre-`Widget::udpate`" and "post-`Widget::update`" stages. This way, `Widget::update` can be used to instantiate children widgets while having the latest parent widget information available, rather than being a frame behind which was causing positional stutter (especially when scrolling) and slow reactions.
- Moved `WidgetMatrix` to the position module and called it `PositionMatrix`. Added a new `WidgetMatrix` type that actually implements `Widget`, and allows for generating a matrix of a single type of `Widget` by automatically positioning/sizing them and delegating each a unique ID.

mitchmindtree:
Now uploaded to crates.io as **0.20.0**.

PR #590, opened 2015-10-16
Fix Graph::pick_widget/bounding_box so that they properly consider widgets' cropped scroll area.

### LINKED ISSUE #446: DropDownList Element doesn't crop properly to scroll area.
DropDownList's `Element` should be cropped when drawn, however the cropping seems to be ignored for some reason.

In both of the following examples, the DropDownList's height should be cropped to the top and bottom of the scrollbar, however it is currently overlapping.

incorrectlycroppeddropdownlist2
incorrectlycroppeddropdownlist1

The DropDownList's `Element` is cropped [here](https://github.com/PistonDevelopers/conrod/blob/master/src/widget/drop_down_list.rs#L654).

Here's the [`Element::crop` method](https://github.com/mitchmindtree/elmesque/blob/b2fcc5394cdfc8168a8136f65c3d6933bb3d4c98/src/element.rs#L142) from the elmesque crate and here's where the [cropping is applied](https://github.com/mitchmindtree/elmesque/blob/b2fcc5394cdfc8168a8136f65c3d6933bb3d4c98/src/element.rs#L430-L462) using gfx's `DrawState::scissor` method.

Cropping seems to work fine in the elmesque example if you uncomment [this line](https://github.com/mitchmindtree/elmesque/blob/master/examples/graphics.rs#L59) so I'm not yet sure what could be causing it to be ignored in the `DropDownList`'s case.

### LINKED ISSUE #572: all_widgets example hangs when selecting "Green" or "Blue" in drop down
I haven't looked at the code yet.

### LINKED ISSUE #589: Fix bug where scroll area cropping isn't considered in several methods.
- [x] Graph::bounding_box
- [x] Graph::pick_widget (has it included, but seems to not be working properly)

mitchmindtree:
Fixes #589 
Fixes #572 
Fixes #446

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.20.1"
+version = "0.20.2"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/src/graph/mod.rs
+++ b/src/graph/mod.rs
@@ -229,8 +229,8 @@ impl Graph {
             .find(|&&visitable| {
                 match visitable {
                     Visitable::Widget(idx) => {
-                        if let Some(&Node::Widget(ref container)) = dag.node_weight(idx) {
-                            if container.rect.is_over(xy) {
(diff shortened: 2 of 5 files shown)


mitchmindtree:
Published as **0.20.2**.

Issue #554, opened 2015-09-02
Change `Widget::update` to return a function for mutating the current State rather than produce a whole new State?


mitchmindtree:
Motivation
-------------

Currently, the signature for `Widget::update` is:
```Rust
fn update<C>(self, args: UpdateArgs<Self, C>) -> Option<Self::State>
    where C: CharacterCache;
```

The idea is that you only have to return a new `State` if something has changed between updates. This works quite efficiently for small widgets, however can be quite expensive for larger widgets.

A possible solution is to instead return an `Option<Box<FnMut(&mut Self::State)>>`. With this change, the `Widget::update` signature might look something like this:

```Rust
fn update<C>(self, args: UpdateArgs<Self, C>) -> Option<Box<FnMut(&mut Self::State)>>
    where C: CharacterCache;
```

This way, instead of returning an entirely new state, we return a function that can mutate the state that already exists, reducing large allocations etc for larger widgets.

Concerns
-------------

It would be nice if we didn't have to `Box` the function that we're returning, however I'm unsure if it's possible to make an API that doesn't require it (I think we'd need HKT or something).

Any suggestions or alternatives for handling this problem in a nicer way are welcome!

bvssvni:
What about calling a closure with a closure? Perhaps we could get rid of the `Box`.

Example: https://github.com/PistonDevelopers/graphics/blob/master/src/graphics.rs#L17

mitchmindtree:
I was originally thinking along these lines too, however the `Self::State` that needs to be mutated also needs to be available immutably (at the moment this is done via the `UpdateArgs`) so that the user can compare the builder args to the previous state and first determine whether or not they need to mutate at all.

Perhaps instead of being passed in via the `UpdateArgs`, the `Self::State` can be passed in as a second argument within some special type that implements two methods: one that produces a `&State<Self::State>` for immutable access and another that takes a `&mut FnMut(&mut Self::State)` closure which it uses for updating the state while also setting a flag so that we can know the `Self::State` has been mutated (and in turn that we have to call `Widget::draw` again to produce a new `Element`).

I think this might look something like:

```Rust
fn update<'a, 'b, C>(self, args: UpdateArgs<'a, 'b, Self, C>, state: &mut StateCell<Self::State>);
    where C: CharacterCache;
```

Maybe the `StateView` might look something like this:

```Rust
pub struct StateCell<T> {
    state: T,
    has_updated: bool,
}

impl<T> StateCell<T> {

[...]

PR #595, opened 2015-10-19
Change Widget::update to mutate the Widget::State rather than return a new one. Move input capturing methods from Widget trait to UiCell.

mitchmindtree:
This change addresses the problem highlighted in #554 where large widgets would be required to allocate an entire new `Widget::State` even for the slightest of changes. This is because the only way of updating the state was to return an entirely new state from the `Widget::update` method.

This PR changes the `Widget::update` method so that it no longer returns an `Option<Widget::State>` but instead provides a way of mutating the state via the `State` wrapper which is given via the `UpdateArgs`. All fields not relating to the unique widget state have been moved out of the `State` type and into a `CommonState` type. This allows us to use the `State` type purely as a "cell"-like wrapper around the `WidgetState`.

`State::view` provides immutable access to the `Widget::State`.
`State::update` allows for mutation of the `Widget::State`. When called, it sets a `has_updated` flag, allowing us to check whether or not the `Widget::State` was updated at all following the call of `Widget::update` and in turn allowing us to know whether or not we need to produce a new `Element` for the `Widget`.

This change also required moving the input capturing methods from the `Widget` trait to the `UiCell`, to be called within the `Widget::update` if necessary.

### DIFF
--- a/src/widget/mod.rs
+++ b/src/widget/mod.rs
@@ -33,89 +33,99 @@ pub mod toggle;
 pub mod xy_pad;
 
 
-/// Arguments for the `Widget::update` method in a struct to simplify the method signature.
+/// Arguments for the [**Widget::update**](./trait.Widget#method.update) method in a struct to
+/// simplify the method signature.
 pub struct UpdateArgs<'a, W, C: 'a> where W: Widget {
-    /// The Widget's unique index.
+    /// The **Widget**'s unique index.
     pub idx: Index,
-    /// The Widget's previous state. Specifically, the state that is common between all widgets.
+    /// The **Widget**'s previous state. Specifically, the state that is common between all widgets,
+    /// such as positioning, floatability, draggability, etc.
     pub prev: &'a CommonState,
-    /// A wrapper around the Widget's unique state, providing methods for both immutably viewing
+    /// A wrapper around the **Widget**'s unique state, providing methods for both immutably viewing
     /// and mutably updating the state.
     ///
     /// We wrap mutation in a method so that we can keep track of whether or not the unique state
-    /// has been updated. If `State::update` is called, we assume that there has been some mutation
-    /// and in turn will produce a new `Element` for the `Widget`. Thus, it is recommended that you
-    /// *only* call `State::update` if you need to update the unique state in some way.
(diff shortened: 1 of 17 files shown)


src/widget/mod.rs
```diff
@@ -714,44 +746,63 @@ fn set_widget<'a, C, W>(widget: W, idx: Index, ui: &mut Ui<C>) where
...
+                // something to do with UpdateArgs' lifetime intersection - I attempted various
+                // other lifetime combinations but failed to get anything working that would
+                // appease the borrow checker. As far as I can see the current intersection should
+                // be fine and safe, however the borrow checker isn't convinced.
+                //
+                // - mindtree.
+                let state_mut_ptr: *mut State<W::State> = &mut state;
+                widget.update(UpdateArgs {
+                    idx: idx,
+                    state: unsafe { ::std::mem::transmute(state_mut_ptr) },
```
mitchmindtree: This really shouldn't be required, however I couldn't find any other way to appease the borrow checker. Perhaps there is something I'm missing? Would be nice if someone could review and see if they can get it working without the unsafe. @bvssvni ?

bvssvni: Checking now.

bvssvni: I've reduced it http://is.gd/Xu6jc0

bvssvni: Here is the solution: http://is.gd/LEoLyJ

You need to add another lifetime `UpdateArgs<'a, 'b: 'a, W, C: 'a>` and pass it in to `pub state: &'a mut State<'b, W::State>`. Notice that the second lifetime `'b` inherits the first `'a` such that `'b` always outlive `'a`.

This error happened because `UpdateArgs` moved, and the compiler only has `'a` to reason about.

mitchmindtree: Woah, great stuff @bvssvni ! I tried multiple lifetimes but didn't consider bounding the second lifetime by the first, though it does make a lot of sense :) thanks! I'll implement the change now

mitchmindtree:
All passing unsafe-less :smile_cat: going to merge!

Published as v**0.21.0**.

PR #599, opened 2015-10-22
Update to latest daggy version. Fix bug in update_depth_order.

mitchmindtree:
update_depth_order was traversing both depth *and* position edges, but this is now fixed.


mitchmindtree:
Uploaded as **0.21.4**

PR #601, opened 2015-10-23
Remove unnecessary code from default_width/height. Clarify comment in custom_widget example.

mitchmindtree:
The common builder is already checked, no need to check it there too.


mitchmindtree:
Uploaded as **0.21.5**.

PR #603, opened 2015-10-23
Fix bug in Element scroll stack folding. Update to latest crop bug fix in elmesque.

mitchmindtree:


mitchmindtree:
Uploaded as **0.21.6**.

PR #618, opened 2015-11-05
First pass at allowing user to move between TextBoxes with keyboard - Don't Merge

psFried:
This is just a first pass at getting this to work. This PR is just to provide a place for people to comment and give feedback on the approach. Do not merge this! 

This allows users to move between TextBox widgets by hitting enter. One problem I see with this approach is that whatever TextBox had previously captured input will try to un-capture after the next TextBox has started capturing. This results in a bunch of warnings printed to stdout.

### DIFF
--- a/src/graph/mod.rs
+++ b/src/graph/mod.rs
@@ -183,7 +183,7 @@ impl Graph {
             floating_deque: Vec::with_capacity(capacity),
         }
     }
-    
+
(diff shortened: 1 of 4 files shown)


psFried:
I am finally getting around to working on this and have figured out a slightly different and simpler approach. Closing this PR for now. I'll re-open another one once the branch is ready.

mitchmindtree:
:+1:

Issue #576, opened 2015-10-01
Conrod should expose (or at least re-export) everything a user needs to use their own graphics and event backends.


mitchmindtree:
Right now conrod is already quite generic in it's Graphics and Events dependencies, however it can seem quite confusing and messy to a user as it requires using various different crates from the Piston eco-system.

I think we could solve this issue quite nicely by exposing and re-exporting everything that a user needs to use their own Graphics and Events backends through conrod itself - perhaps via a `backend` module, which re-exports (and inline documents) the `Graphics` and `Event` traits. We should also provide a demonstration of how to implement these manually, perhaps with a basic example of implementing them directly on `glutin` and `glium`.

Really, a user should not need to import any crates besides conrod itself when they desire to use conrod within their own project.

We quite often get feedback critiquing the complexity of the eco-system that surrounds conrod, i.e. just now in IRC:

> OmniMancer - I think having an  interface that doesnt require even indirectly depedning on piston would possibly be helpful

> jpernst - yeah, i evaluated conrod for use on top of my framework but decided against it because it pulls in too much

bvssvni:
This is a good idea. It could help people looking for using Conrod in this way. For example, if you are developing your own framework and want an UI.

bvssvni:
To address the confusion problem, I've written an overview of Piston https://github.com/PistonDevelopers/piston/wiki/Piston-overview.

mitchmindtree:
Nice one, the overview looks great :+1:

PR #626, opened 2015-11-15
[Breaking Change] Replace elmesque crate with primitive graphics widgets (see comment for more details).

### LINKED ISSUE #421: Add `widget_node_capacity` and `widget_edge_capacity` builder methods to `Ui`.
For the case that someone wants finer grained control over the amount of memory to allocate upon initialising the widget `Graph`'s Node and Edge `Vec`s.

### LINKED ISSUE #466: Multi-line labels
It'd be nice to have configurable wrapping for Label or some other widgets.

### LINKED ISSUE #506: Need some way of adding unique styling for custom Widgets to `Theme`.
I was thinking we could add a `HashMap<&'static, Box<Any + Encodable + Decodable>>` (or maybe [AnyMap](https://github.com/chris-morgan/anymap)) to the `Theme`, where `&'static` is the `Widget`'s `unique_kind` and the boxed data is the styling, however I'm unsure if it is possible to serialise dynamic data like that.

For now the internal widgets just have their styling hard coded into the struct (this is of course not possible for custom widgets).

### LINKED ISSUE #509: Allow for Widgets to be `set` within an applications `update` stage (and allow for arbitrary update rates).
I've had a few people raise the concern that it feels strange and non-idiomatic to update all widgets every `draw` aka `render` stage, which I totally understand.

Following all of the recent changes within conrod, i'm realising that it should be very easy to allow a user to instead call all of the widgets within the `update` stage, and then  only call `ui.draw` in the draw stage. This feels a lot more idiomatic while still retaining the current immediate-mode ease of use. It will allow users to set arbitrary update rates etc. It will also allow us to add methods along the lines of `only_update/redraw_on_input`, which would allow a user to say that they only want the UI to update/redraw if being interacted with. Atm we're redrawing every frame which is hands down conrod's biggest bottleneck right now, but with this change conrod would be almost free unless being interacted with.

- [x] Remove error warning about multiple updates per render.
- [x] Review `handle_event` (might need to move some stuff from render event to update).
- [ ] Add something like `Ui::only_update/redraw_on_input`.
- [x] Add a method for returning the `Ui`'s drawable `Element`. This could be an alternative to `Ui::draw` and would allow for conrod to run on a separate thread, etc.
- [ ] Change examples to set widgets in update stage.

### LINKED ISSUE #522: Separate out the text logic into Elmesque crate
In order to improve Conrod's text support in the future (e.g. multi-line text), all of the text logic should be gathered together to be shared with all of the widgets.

### LINKED ISSUE #555: Change from using `Option` to `std::iter::once` when chaining together `Form` Iterators.
I've used `Option` throughout the widget `draw` methods to chain together `Form` `Iterator`s (in order to avoid allocations), however I think the more idiomatic way would be to use `std::iter::once` (didn't see it until today).

For example, the line [here](https://github.com/PistonDevelopers/conrod/blob/master/src/widget/envelope_editor.rs#L656)
```Rust
Some(circle_form).into_iter().chain(Some(text_form))
```
would become
```Rust
once(circle_form).chain(once(text_form))
```

### LINKED ISSUE #581: Add a `.size_of(WidgetId)` method that sets the dimensions for a Widget from the dimensions of the Widget with the given ID.

### LINKED ISSUE #582: Add an internal, transparent `window` widget to simplify placing Widgets relative to parts of the window.
The `Ui` would instantiate this upon its construction and resize its dimensions whenever the window was resized. Its position would always be (0, 0) (centred).

The `Ui` would expose a method for retrieving its widget index so that it can easily be used with positioning and sizing methods.

Whenever a background or clear color is given, we'd simply set this as the color of this `Window` widget.

### LINKED ISSUE #616: Replace elmesque with a conrod-specific layout tree?
@bvssvni has been doing some bench-marking and it led to some discussion about alternatives to elmesque's purely-functional allocation-heavy graphics layout approach.

[Original discussion here](https://github.com/PistonDevelopers/turbine/issues/31#issuecomment-153954776).

> This would be quite the breaking change, however there could be a number of benefits to this approach:
> -  The design would not be restricted to elmesque's aims to stay representative of the elm-lang graphics modules.
> - We could remove the confusing distinction between the `Form` and `Element` types and consolidate them under one `Form` type for example.
> - It could be easier for a Widget to re-use and update its existing graphics description/layout (elmesque doesn't provide a way to do this).
> - It would be easier to provide a simple conrod backend API (PistonDevelopers/conrod#576) as we could easily make the rendering of our custom system generic over our own backend (this might be trickier or not as nice using elmesque).
> - One less dependency - slightly easier to maintain.

> I can picture this being done as a "graphics layout tree", where rather than building the tree using boxed recursion, we provide a little tree data structure (a wrapper around something like [rose_tree](https://crates.io/crates/rose_tree)) to each widget which only gets allocated once and can be re-used every `Widget::draw`. The API could be very similar to as it is now, just with the tree being described in the data structure rather than the type recursion, removing the need for all the elmesque related allocations. We might even be able to do all widgets within a single tree - rather than providing each widget with it's own unique tree, we could provide a safe wrapper around the branch of the master tree that represents that widget's graphics layout.

### LINKED ISSUE #617: Remove the need for `Form` and `Element` by representing graphics fundamentals as `Widget`s and taking advantage of widget layout.
This seems so obvious in hindsight!

See original idea [here](https://github.com/PistonDevelopers/conrod/issues/616#issuecomment-153973156).

Edit: I think the reason this wasn't obvious before is because conrod's layout only recently became very powerful.

### LINKED ISSUE #628: Envelop editor causes crash when all points are removed by right click
Removing all the points in the envelop editor causes the program to crash.

     Running `target/debug/examples/all_widgets`
     thread '<main>' panicked at 'index out of bounds: the len is 0 but the index is 0', src/widget/envelope_editor.rs:633
     Process didn't exit successfully: `target/debug/examples/all_widgets` (exit code: 101)

### LINKED ISSUE #629: Clarify the "widget setting stage" by passing the `Ui` a function.
In order to achieve #509 we need to clarify the stage in which `Widget`s are `.set` by the user. For the purpose of this issue, we'll call this the "widget setting stage".

At the moment, the `Ui` simply assumes that the widget setting stage is the entire period between calls to `.draw` (or `.draw_if_changed`). This means we're currently locked to the rate at which the user `.draw`s the `Ui` and we're only able to do clean up and prep at the beginning and end of `.draw`.

The nicest solution I can think of is to add a `.set_widgets` method to the `Ui` that takes a function during which the user sets all their `Widget`s.

For example, the `counter.rs` example could change from this:

```rust
    // Poll events from the window.
    for event in window {
        ui.handle_event(&event);
        event.draw_2d(|c, g| {

            // Generate the ID for the Button COUNTER.
            widget_ids!(CANVAS, COUNTER);

            // Create a background canvas upon which we'll place the button.
            conrod::Split::new(CANVAS).pad(40.0).set(&mut ui);

            // Draw the button and increment `count` if pressed.
            conrod::Button::new()
                .middle_of(CANVAS)
                .dimensions(80.0, 80.0)
                .label(&count.to_string())
                .react(|| count += 1)
                .set(COUNTER, &mut ui);

            // Draw our Ui!
            ui.draw_if_changed(c, g);
        });
    }
```

to this

```rust
    // Poll events from the window.
    for event in window {
        ui.handle_event(&event);
        event.update(|_| ui.set_widgets(|ui| {

            // Generate the ID for the Button COUNTER.
            widget_ids!(CANVAS, COUNTER);

            // Create a background canvas upon which we'll place the button.
            conrod::Split::new(CANVAS).pad(40.0).set(ui);

            // Draw the button and increment `count` if pressed.
            conrod::Button::new()
                .middle_of(CANVAS)
                .dimensions(80.0, 80.0)
                .label(&count.to_string())
                .react(|| count += 1)
                .set(COUNTER, ui);
        }));
        event.draw_2d(|c, g| ui.draw_if_changed(c, g));
    }
```

This gives us more control over the widget setting stage, allowing us to do preparation and cleanup directly before and after the given function is called. It also lets us more clearly distinguish between the "update" and "draw" stages.

I think I might tackle this within #626 as it should help to resolve some scrolling issues I'm having (the issues are related to widgets that are conditionally drawn - something which should be easier to track having a clearer distinction between stages).

mitchmindtree:
~~This is a WIP - **Do not merge** just yet.~~

This is quite a hefty change, and in turn is leaking into quite a few areas:

- Add the widget::primitive module, hosting all primitive widgets. This module should contain all the primitive widgets a widget designer could possibly need to implement their own complex widgets.
- Added an example demonstrating each of the primitive widgets.
- `Widget`s will now compose themselves of primitive graphics widgets rather than drawing their own `Element`s or `Form`s. This allows us to remove the elmesque crate and use our own graphics layout instead, in turn providing a far simpler (and more efficient) graphics backend.
- `Theme` will now support custom widget styling (rather than hard-coded style types).
- Added methods to the `Sizeable` trait that allow setting a `Widget`'s dimensions to match the dimensions of some other `Widget`.
- Added a `TitleBar` widget, which `Canvas` will now use instead of its own custom `TitleBar` type.
- Added a `Widget::picking_passthrough(bool)` method which allows a user to indicate whether their widget should be pick-able by the mouse or other widget picking methods. This is useful for primitive graphics widgets that are only designed to be used as graphical elements, and that are not themselves interactive.
- Fixed a bug in `Graph::scroll_offset` where the behaviour would be incorrect if two widgets shared both `Edge::Depth` and `Edge::Position`.
- Removed `Label` widget in favour of new `Text` primitive widget that handles automatic line-wrapping, line-spacing, etc.
- Moved color.rs module from elmesque into conrod.
- Removes rustc-serialize as the first step towards using serde (perhaps as a feature).

All-in-all I think this should be a massive API improvement, and allow conrod to be a lot simpler to work with regarding graphics (elmesque has been a common point of confusion). It should also be much faster than elmesque as drawing will no longer require near as many allocations.

Closes #421.
Closes #466.
Closes #506.
Closes #509.
Closes #522.
Closes #555.
Makes a start on #576.
Closes #581.
Closes #582.
Closes #616.
Closes #617.
Fixes #628.
Closes #629.


mitchmindtree:
The `Theme` serialisation will also need a re-do due to the addition of handling 3rd party `Widget::Style`s.

Perhaps this is an opportunity to move away from the **rustc_serialize** crate in favour of **serde**.

bvssvni:
I think moving to serde now is a good idea.

pyros2097:
Nice! Me like performance! Hulk Smash!

mitchmindtree:
One issue that I've come across regarding the use of `Widget`s as primitive graphical elements of other `Widget`s is that it is easy to make some slightly different assumptions about the behaviour of their relationships, compared to regular widget relationships.

Say that we have ***a -> b*** where our child ***b*** is a `Widget` being used as a graphical element of the parent `Widget` ***a***. In this case, we'll say that ***b*** is a `Rectangle` widget being used as a graphical element for a `Canvas` (a container kind of widget) ***a***. This allows:
- Use of the `Positionable` methods to position ***b*** relative to ***a***, i.e. `.middle_of(canvas_idx)`.
- Use of the `Sizeable` methods to size ***b*** relative to ***a***, i.e. `.dim_of(canvas_idx)`.
- ***a*** doesn't need a draw method, as ***b*** takes care of the drawing.
- ***a*** is the implied depth parent of ***b*** as ***b*** was placed upon ***a*** using the `.middle_of` method.

There are a few issues that are easy to overlook here:

1. If our `Canvas` ***a*** is instantiated as scrollable by a user, then our `Rectangle` ***b***'s position will be offset by ***a***'s scrolling, because we described ***b*** as a depth child of ***a***... a disaster! What we actually want is for ***b*** to stay fixed to ***a*** (even though it is a depth child of ***a***) as ***b*** is actually a *graphical element* of ***a*** and not just another child widget.
2. When a user moves their mouse over our `Canvas` ***a***, the internal `pick_widget` method always picks ***b*** instead of ***a***, as ***b*** covers the entirety of ***a***'s area... another disaster! What we actually want is for ***b*** to be considered a *graphical element* of ***a***, so that whenever `pick_widget` picks ***b*** at the top, it actually returns the index for ***a***.
3. Say we want to place a `TitleBar` widget ***c*** at the top of our `Canvas` ***a***. We could say that, if our `Canvas` ***a*** has some `TitleBar` ***c***, then we want its `kid_area` to be the remaining area directly under the `TitleBar` ***c***. This means that widgets placed upon the `Canvas` ***a*** will be correctly placed in the area under the `TitleBar` ***c***. Now, in ***a***'s `Widget::update`, we need to actually intantiate and position our `TitleBar` ***c***. Intuitively we want to write `title_bar.mid_top_of(canvas_idx).width_of(canvas_idx)`, which should make sense. However, remember that ***c*** is a depth child of ***a***, meaning that when we say `.mid_top_of`, we're actually saying that we want to position it at the top of ***a***'s `kid_area`, and not the *actual* top of ***a*** as we assumed... disaster! We need conrod to know that ***c*** is actually a *graphical element* of ***a***, and that this implies it should actually be positioned upon ***a***'s total area.

I think this clearly suggests that we need some way of describing that a widget can be a graphical element of some other widget. The question is, what is the ideal way of representing this within conrod? I think there are a few options:
- The first solution that I thought of was to have each `Node::Widget` contain its own internal `Graph` allowing for each widget to have its own complex graphics widgets layout that is clearly a part of the widget itself. Despite this pretty clearly solving the above issues, I think it might introduce a fair bit of unnecessary complexity. It adds another dimension of traversal by adding the addition of `Graph`s within `Graph`s. We would also likely lose cache coherence; at the moment all widgets sit within the `Graph`'s single contiguous node slice, however this solution would require allocating a new slice for every widget that has its own group of graphical elements (this would be the case for most widgets), likely slowing down graph traversal considerably.
- I then thought it might be possible to instead add unique fields that describe our desired behaviour to the widget `Container` stored within each `Node::Widget` within the `Graph`. For example, there could be:
    - `attach_scroll_offset_to` (`None` by default): when `Some(other_idx)` would describe that a widget should collect its `scroll_offset` via some other widget's depth position within the graph rather than its own. This would allow us to attach a child graphics widget to some parent, solving the problem described in the 1st issue above.
    - `picking_passthrough` (`false` by default): describing whether or not a widget should be ignored within the picking algorithms. This would fix the 2nd issue.
    - `place_on_kid_area` (`true` by default): describing whether or not a widget should be place upon the parent's `kid_area`. If `false`, it would instead be placed upon the parent's total area. This would fix the 3rd issue.

    This is nice in that we could offer a user control over the behaviour of each individual issue described above. My concern is that the majority of the time, the behaviour of all of these are related to a single question, "is the widget a graphical element for some other widget or not?". Thus perhaps instead of these 3 fields we could add just one; `graphical_element_of` which is `None` by default, and `Some(idx)` where idx is the index of the widget for which our widget is a graphical element. This brings upon the realisation that, what we are actually describing here is a relationship between two separate widgets. This leads to the next possible solution...
- Add a `Graphic` variant to the `Edge` enum, which describes the exact relationship I mentioned at the start of this comment: ***a -> b*** where our child ***b*** is a `Widget` being used as a graphical element of the parent `Widget` ***a***. I think this is currently my favourite option, as it seems to be the simplest, clearest way of describing the graphical relationship between two widgets while still using the same graph and in turn, the same contiguous slice of memory.

API-wise, there are a couple of options for allowing the user to describe this relationship:
- we could provide a `.graphics_for(other_idx)` builder method on the `Widget` trait, which would allow a user to specify that the widget they're currently instantiating is simply a graphical element of the widget at the given index (I think this is my preferred option).
- add an alternative to `.set(idx, &mut ui)`, such as `.set_graphics_for(other_idx, idx, &mut ui)`.
- always assume that primitive widgets are graphical elements of their parent widgets? This might be the simplest option, however it probably isn't a safe assumption to make as I can already imagine situations in which a user might expect a primitive to behave just like a regular widget.

masche842:
@mitchmindtree I like where this is going! As mentioned in #616 I played around with this today.
It has been easy to set up a custom backend for piston2d-graphics and I think I'm fine with this dependency (for me it will only behave as an interface anyway as I can't use one of the implemented backends). 
But there's a thing that confuses me: graphics already defines methods for rectangles, ellipses, etc.. But in your current revision (861726a) you do reimplement them (for the ellipses even with fixed resolution). Is this just an interim stage because elmesque doesn't offer the means for describing theses as `Element`?
This leads to my second question: I then tried to get rid of elmesque to get a feeling for what has to be done. I ended up calling widget::draw in ui::draw, assembling DrawState from Cache. This does the job but feels kind of messy. Do you already have an architecture in mind that minimizes changes in widgets?

I'm really looking forward to that pull request being finished & merged, would be a pleasure to help!

bvssvni:
@mitchmindtree Thanks for the info! It was clear and well written. I think adding a `Graphics` edge variant sounds good.

mitchmindtree:
Sorry about the wait everyone, will hopefully have this done in the next day or two! :pray:

bvssvni:
@mitchmindtree No worries! Take your time.

mitchmindtree:
As I had to re-implement the text drawing, I thought I might as well have a go at getting auto-wrapping multi-line text going - here's a WIP:

screen shot 2015-12-02 at 1 43 12 pm

Still some tweaking to go, it will also support left-aligned, centred and right-aligned text.

pyros2097:
Nice! Could this be possibly used to create a fast Text Editor for Rust or will the Immediate mode GUI make it more slow and memory intensive?

mitchmindtree:
@pyros2097 it should be! Conrod is all retained under the hood (it re-uses allocations, strings, textures etc like regular GUIs) and provides ways for drawing lazily, so I don't see why we shouldn't be able to reach speeds close to that of regular retained GUIs, but we'll see :sweat_smile:

bvssvni:
@mitchmindtree This is going to be the most epic PR ever!

bvssvni:
This PR needs rebasing.

mitchmindtree:
@bvssvni just rebased :+1: I went away for the weekend but now I'm back to work on this and only have a couple more widgets to update :smile_cat:

masche842:
Wow, your progress on this is amazing! I'm so looking forward to it...
I should have ask earlier, but, do you need help with anything?

If this pull request is finished (or close to be finished) I will implement my own backend anyway. Shall I try to provide an example for, say, an headless backend?

mitchmindtree:
Thanks @masche842 !

Yeah this PR is pretty close to finished now I think (just need to update `EnvelopeEditor` and `TextBox` and re-implement draw_if_changed).

I'm curious about what you mean by headless backend?

Btw, I forgot to reply to this!
> But there's a thing that confuses me: graphics already defines methods for rectangles, ellipses, etc.. But in your current revision (861726a) you do reimplement them (for the ellipses even with fixed resolution). Is this just an interim stage because elmesque doesn't offer the means for describing these primitives as Element?

The reason I tried to use as little as possible from the graphics crate was in case we want to write a more minimal conrod-specific `Graphics` backend trait (with only what we need) sometime after this PR. We could still use piston backends by doing a blanket impl for the piston-graphics trait. I'm unsure just how much more we could simplify it (haven't put a great deal of thought into it just yet) but it's an often requested change so will be worth putting some thought into it.

Re ellipse with fixed resolution: I'm happy for us to add a resolution param :)

Btw, I've also been contemplating moving some of the widget fields to type params (at least all the ones that users only really need to set once and do not require dynamic control over) but only if we can retain a nice API. It might require waiting for [this](https://github.com/rust-lang/rust/issues/27336) to land in stable, but once it does, I think we could achieve it while retaining the same builder-style API. Anyways, this is probably best for another issue and PR heh :)

masche842:
@mitchmindtree  

> I'm curious about what you mean by headless backend?
> 
> ...in case we want to write a more minimal conrod-specific Graphics backend trait (with only what we need) sometime after this PR. We could still use piston backends by doing a blanket impl for the piston-graphics trait...

Oh, I was one step ahead of this pull request then :-)
I just meant exactly that: Providing a minimal Graphics backend example as a guide/documentation. 
I kind of forgot that extracting a conrod-specific backend trait is not part of this PR. 
And it really shouldn't be, it's already huge without.

As @bvssvni  suggested in #issuecomment-157450758 I will implement my own backend in piston-graphics first and see how it will work out. 
But with the elmesque dependency out of the way and a paved road to Graphics backend abstraction it's tempting to get rid of the piston-graphics dependency altogether...

mitchmindtree:
@masche842 That would be *awesome* if you do get a chance (the minimalised `Graphics` abstraction, that is)! I think a lot of people would appreciate it. That way we could possibly make piston-graphics an optional default cargo feature instead, so that by default it's still easy to get started using piston, but there's an option to not use it whatsoever if that is what the user requires :+1:

mitchmindtree:
Should be fiiiiinally ready :tada:

bvssvni:
Wow!

mitchmindtree:
About to head to sleep but will merge and publish this when I wake up if no one beats me to it (or has any issues) :+1:

mitchmindtree:
Published as **0.24.0**!

ebfull:
Awesome work! :tada:

masche842:
Thanks @mitchmindtree for this huge and awesome PR! 
I think conrod profits and will profit a lot from this!  :+1: :+1: :+1:

As discussed before I added a sketch for a simple backend implementation here: #658

sjmackenzie:
Epic

Issue #631, opened 2015-12-02
clock_ticks doesn't compile


kookerus:
The clock_ticks crate doesn't compile and is deprecated. The dep should be changed to the time crate.

PR #632, opened 2015-12-02
modified Cargo.toml to use the "time" crate instead of the deprecated "clock_ticks" module

kookerus:
"clock_ticks" is deprecated and broken, the git repo says to use "time" instead.


kookerus:
This commit fixes issue #631

kookerus:
whoops, I forgot to change a few things

PR #640, opened 2015-12-13
Fix Panic When TextBox is Empty

iwillspeak:
Previously if all the characters were deleted from a text box then we
would try to calculate the height based on 0 lines of text. This causes
the thread to panic with an arithmetic overflow on OS X.

This commit fixes things by rendering a text box with 0 characters in it
as the height of a single line.

### DIFF
--- a/src/widget/primitive/text.rs
+++ b/src/widget/primitive/text.rs
@@ -1,4 +1,4 @@
-
+use std::cmp;
 use {
     Align,
     CharacterCache,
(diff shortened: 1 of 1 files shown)


bvssvni:
Tested at OSX. Merging.

mitchmindtree:
Thanks for this :)

iwillspeak:
No worries. Glad I could help. Really liking the look of Piston so far :+1:

PR #654, opened 2015-12-22
Add new Positionable methods with support for margining. Removal of Split in favour of moving functionality to Canvas.

mitchmindtree:
Widgets can now be offset from the edge along which they are placed using the `margin` variant of some `Positionable` methods. I.e. we now have `.mid_left_with_margin(scalar)` to match `.mid_left()` and `.padded_w_of(idx, scalar)` to match `.w_of(idx)`. This has allowed us to remove all uses of absolute coordinates from all of the examples :smile_cat:

Also renamed various `Sizeable` methods to follow the conventions of `Positionable` and updated some `Ui` methods to match.

The `Split` type (often a point of confusion due to its name) has been removed in favour of moving all of its functionality onto the `Canvas` type. This wasn't possible when `Canvas` was originally conceived as we never used to be able to instantiate widgets inside other widgets, however now that we can, `Split` is no longer necessary.

Canvas split lengths can now also be set by "weight"s (as an alternative to giving an absolute length).


mitchmindtree:
Going to merge as travis' failure seems to be due to a regression on nightly (it's building fine on stable).

mitchmindtree:
Published as **0.27.0**.

PR #662, opened 2016-01-08
Scrolling refactoring and Bug fixes. Add `Axis` enum arg to `Edge::Position` in graph.

### LINKED ISSUE #659: Refactor Scrolling (again).
Problem
-----------

At the moment, the job of a `Widget`'s `KidArea` is a little ambiguous. It has progressively undertaken multiple roles:
- The place upon which that `Widget`'s children `Widget`s are placed when using placement `Positionable` methods, i.e. `.mid_top()`, `.bottom_right()`.
- The area which is cropped when scrolling children `Widget`s.
- The area upon which `Widget`s are offset when scrolled.

This has slowly introduced complexity into the widget positioning logic, especially with relation to scrolling, which has lead to some bugs that I've noticed in my own personal projects that are extremely difficult to reason about. Thus, it is clear that the scrolling logic needs to be refactored into a more palatable, clearly defined, robust, step by step process.

Solution
-----------

*It's quite difficult to find thorough writing on the topic of handling scrolling logic, and even harder for immediate-mode interfacing GUIs. The following draws some inspiration from servo's [scrolling overview](https://github.com/servo/servo/blob/master/components/compositing/scrolling.md), however only loosely as servo is retained and event based, whereas conrod is immediately reactive so requires more fine-grained, step-by-step details to be covered.*

1. First, the name of the opt-in method for scrolling should probably change from `.scrolling(bool)` to `.scroll_kids()` to clarify that opting-in to scrolling on the widget *Foo* means that *Foo*'s *children* widgets are offset by scrolling, where *Foo*'s `KidArea` is the root rectangle.

2. Rather than piling multiple roles onto `KidArea`, the roles should be split up into separate types and stages:

    1. The `KidArea`defines the `Rect` upon which children widgets may be placed. *Foo*'s `KidArea` `Rect` does not change with respect to *Foo*'s scrolling and is simply used for handling the placement of child widgets specified by their *continuous positioning arguments*. By default, this is the `Widget`'s entire bounding `Rect`, however it can be specified by overriding the `Widget::kid_area` method.

    2. For scrollable widgets, produce a `scrollable_rect` (a `scrollable_range` for each axis) at the beginning of the scroll calculation stage during the `set_widget` function. This rect specifies the total area which may be "offset" from the ***root*** rectangle (the `KidArea`). The `scrollable_rect` will be determined by the bounding box around both the `KidArea` and all ***un-scrolled*** ***visible*** children widgets. Note that this means when reading all children widgets' `Rect`s to construct the bounding `Rect`, we must first subtract the scrollable parent widget's ***current*** `scroll_offset` from their absolute `xy` positions in order to get their ***original*** aka ***un-scrolled*** position. Thus, the calculated `scrollable_rect` should be identical each update regardless of changing scroll offset. It should only change if:
        - One or more of the child widgets are manually moved by the user, changing the dimensions of the bounding box.
        - One or more new child widgets appear, expanding the size of the bounding box.
        - One or more child widgets disappear, reducing the size of the bounding box.

    3. Use the calculated `scrollable_rect` to determine the min and max scroll_offset bounds for each axis, where:
        - The `min` bound of an axis is the distance between the start of the scrollable_rect range and the start of the `KidArea` rect range.
        - The `max` bound of an axis is the distance between the end of the `KidArea` rect range and the end of the scrollable_rect_range.

    4. Get the `current_scroll_offset`.
        - If this is not the first time the widget has been instantiated, this value will be the widget's currently stored `scroll_offset`.
        - If this is the first time the widget is instantiated, this will either be `0`, or perhaps some other value specified by a given `initial_scroll_alignment`.

    5. Produce the `additional_scroll_offset` by summing:
        - Scrolling of the mouse wheel.
        - Movement of the scrollbar.

    6. Determine the `new_scroll_offset` by summing the `additional_scroll_offset` onto the `current_scroll_offset`, clamping the result to the previously calculated `min` and `max` bounds. This `new_scroll_offset` will then be stored as the widget's new `scroll_offset`, which children widgets will then sum onto their own positions during the `calc_xy` function.

3. Introduce a `.crop_kids()` method, which crops the children widgets of some widget to its `KidArea`. The `.scroll_kids()` method will then call this internally, rather than having its own special-case cropping as is currently the case. This will allow the user more flexibility, being able to crop children widgets whether or not the parent is scrollable.

### LINKED ISSUE #660: Must add Edge::Position for *both* axes.
At the moment, an `Edge::Position` is only added between the widget which we're relatively placed to along the X axis, even if we're also relatively positioned to a widget on the Y axis. We need to add an axis field to the `Edge::Position` variant - something like `pub enum Axis { X, Y }`.

This is currently causing buggy behaviour in scrolling.

### LINKED ISSUE #661: crates.io not updated to latest version
Looks like the color definitions have not be updated in the 0.28.0 version of the crate. Getting an undefined error with the following code:

```
extern crate conrod;

fn main() {
    println!("{:?}", conrod::color::BLUE);
}
```

Replacing the crate with the github link  works.

Checking the multirust cargo src shows conrod version at 0.28.0 with the old color definitions. I even tried uninstalling rust and clearing all the cached cargo data.

mitchmindtree:
~~**WIP - Do not merge yet**~~

Closes #659.
Fixes #660.
Closes #661.

Ok, refactoring complete!

I'm much happier with these changes, they're much clearer and I can confidently say that I understand the behaviour of each distinct step of the process - not something I can say for the old scrolling setup, which was a little hack-ish and resulted in some hard-to-track-down bugs.

This also fixes bugs in the graph logic related to the relative positioning of widgets. Previously, the graph only had a single edge to represent relative positioning across both axes. I've added an `Axis` enum to the `Position` variant, which allows us to track relative positioning uniquely across each axis. This allows for saying cool stuff like `.align_left_of(X).down_from(Y)`, and to know that relative positioning logic will be handled properly when evaluating scroll offsets, etc.

Finally, scrolling will now also properly consider the `Padding` of the `KidArea` of the widget that has `scroll_kids` enabled. I.e., you can now scroll past the edge of the last widget by the padding amount that is specified for that edge. This is just like how there's a little bit of white padding when you scroll past the widgets at the bottom of this github page.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.28.0"
+version = "0.29.0"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -173,7 +173,7 @@ fn main() {
 fn set_widgets(ui: &mut Ui, app: &mut DemoApp) {
 
     // We can use this `Canvas` as a parent Widget upon which we can place other widgets.
-    Canvas::new().frame(app.frame_width).color(app.bg_color).scrolling(true).set(CANVAS, ui);
+    Canvas::new().frame(app.frame_width).color(app.bg_color).scroll_kids().set(CANVAS, ui);
 
     // Text example.
     Text::new("Widget Demonstration")
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -37,6 +37,7 @@ use conrod::{
 };
 use conrod::color::{self, rgb};
 use piston_window::{EventLoop, Glyphs, PistonWindow, UpdateEvent, WindowSettings};
+use std::sync::mpsc;
 
 
 type Ui = conrod::Ui<Glyphs>;
@@ -76,12 +77,16 @@ struct DemoApp {
     circle_pos: Point,
     /// Envelope for demonstration of EnvelopeEditor.
     envelopes: Vec<(Vec<Point>, String)>,
+    /// A channel for sending results from the `WidgetMatrix`.
+    elem_sender: mpsc::Sender<(usize, usize, bool)>,
+    elem_receiver: mpsc::Receiver<(usize, usize, bool)>,
 }
 
 impl DemoApp {
 
     /// Constructor for the Demonstration Application model.
     fn new() -> DemoApp {
+        let (elem_sender, elem_receiver) = mpsc::channel();
         DemoApp {
             bg_color: rgb(0.2, 0.35, 0.45),
             show_button: false,
@@ -114,6 +119,8 @@ impl DemoApp {
                                    [0.3, 0.2],
                                    [0.6, 0.6],
                                    [1.0, 0.0], ], "Envelope B".to_string())],
+            elem_sender: elem_sender,
+            elem_receiver: elem_receiver,
         }
     }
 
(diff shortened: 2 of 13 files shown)


mitchmindtree:
I've udpated the example to avoid the nightly regression [here](https://github.com/rust-lang/rust/issues/30519#issuecomment-166754771) (the regression was considered a bug fix), however it's hard to tell whether or not this is now fixed, as it looks like travis is now failing due to [*another* regression on nightly](https://github.com/rust-lang/rust/issues/30713). It looks like [there's a fix for it](https://github.com/rust-lang/rust/pull/30733) which should land soon, but there's no point in having it block this as we're still building fine on stable.

mitchmindtree:
Just updated the main comment with more thorough details of this refactoring.

Travis is passing on stable and I've successfully tested the changes with the all_widgets.rs and canvas.rs examples, as well as on two personal projects with much more intricate, nested scrolling behaviour.

I've published these changes as **0.29.0** - this also includes the changes mentioned in #661 and closes that issue.

Issue #569, opened 2015-09-29
Store all input `Event`s between updates so that they can be inspected by `Widget`s


mitchmindtree:
Right now, `Widget`s can only check the present state of the `Mouse`. It would probably be more useful to store each of the input events that have occurred between each update in a single buffer along with methods for quickly filtering through it in useful ways.

This would also allow for fixing #384 .

bvssvni:
Is it possible to store `E: GenericEvent`?

I guess this will be a problem if people are calling Conrod with two different types of events.

bvssvni:
We could map events back to `Input` and store them in a `Vec<Input>`.

mitchmindtree:
@bvssvni yeah I was imagining a `Vec<Input>` :)

Perhaps in a wrapper type that offers a collection of methods yielding different useful iterators over the kinds of input?

bvssvni:
Can we do this in a loop in `set_widget`? It seems possible to call `Widget::update` multiple times.

mitchmindtree:
@bvssvni hmm I'm not sure what you mean?

I think it might be a bit too expensive to call `Widget::update` multiple times in a loop, as some widgets might have complex `update`s. Then again, I guess it would hardly ever get called more than once as it would be rare to have more than one or two `Input` events between `update`s.

I imagined the `Ui` would store the `Vec<Input>` and each time `Ui::handle_event` is called with an `Input`, it would add the event to the stack. The `Vec<Input>` would be reset between updates.

The widgets would receive a view of this `Vec<Input>` (perhaps via the wrapper type providing helper methods as described in my previous comment) within the `UserInput` type that is given via the `UpdateArgs` when `Widget::update` is called. That way each time a widget's update is called, it just handles the events that happened since its last update.

bvssvni:
The behavior of a widget should not depend on which partition of events it received. For example, if the widget capturing the mouse is determined in `set_widget`, then it can't change in `Widget::update`.

mitchmindtree:
hmm I'm still not following? The widget that's capturing the mouse *is* currently determined within the `Widget::update`, i.e. [here](https://github.com/PistonDevelopers/conrod/blob/master/src/widget/button.rs#L151).

bvssvni:
Ok.

bvssvni:
@mitchmindtree I am trying to get up to speed on the source. Are you on IRC?

mitchmindtree:
[...]

PR #663, opened 2016-01-08
Mouse input refactoring

psFried:
This is the beginning of refactoring mouse input handling. The basic idea here is to simplify how widgets handle mouse input in most cases. A new mouse::events module was added that provides higher level event abstractions over low-level mouse events. Widgets can just call `Mouse::events()`, which returns an iterator over `MouseEvent`s. MouseEvents are simple things like Click, Drag, or Scroll. 

Part of the goal here is to obviate the need for widgets to 'capture' the mouse, except in a few rare cases. Turns out, having widgets making calls to capture/uncapture inputs makes it really difficult to add methods to the Ui that control which widgets take focus. Tabbing between text boxes, as well as having external sources capture inputs should get a lot easier once actually capturing input becomes unnecessary in most cases.

There is one breaking change here. Previously, the `Mouse` struct had four fields for it's buttons: `left`, `right`, `middle`, and `unknown`. I changed that to instead use a `ButtonMap`, which maps `input::MouseButton` enums to `ButtonState` structs. Where before, one could call `mouse.left` to get the state of the left mouse button, now you would call `mouse.buttons.get(MouseButton::Left)`. This seems a little more verbose, but it allows for all the mouse buttons to work and makes for some cleaner code when updating the state of the mouse. 

I refactored the TextBox widget to use the new mouse events. The rest of the widgets should follow shortly. I made sure to leave all the other state of the mouse visible. Eventually, we can talk about cleaning up some of the fields on `ButtonState`, but that's a discussion for after all the existing widgets are updated to use the new events.

### DIFF
--- a/src/mouse/events.rs
+++ b/src/mouse/events.rs
@@ -15,6 +15,9 @@ use position::Scalar;
 /// just use these events
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum MouseEvent {
+    /// When the mouse button has been pressed but not yet released or dragged.
+    /// This event is created as soon as the button is pressed.
+    Down(MouseButtonDown),
     /// Indicates that the mouse was clicked. A Click event is created when the mouse button is released, not depressed
     Click(MouseClick),
     /// Drag event is created when the mouse was moved over a certain threshold while a button was depressed
@@ -102,6 +105,7 @@ impl MouseEvent {
         use self::MouseEvent::*;
 
         match self {
+            &Down(button_down) => Down(button_down.relative_to(xy)),
             &Click(mouse_click) => Click(mouse_click.relative_to(xy)),
             &Drag(mouse_drag) => Drag(mouse_drag.relative_to(xy)),
             &Scroll(scroll_info) => Scroll(scroll_info)
--- a/src/mouse/events.rs
+++ b/src/mouse/events.rs
@@ -14,7 +14,7 @@ use position::Scalar;
 /// Used for simplified mouse event handling. Most widgets can probably
 /// just use these events
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub enum SimpleMouseEvent {
+pub enum MouseEvent {
     /// Indicates that the mouse was clicked. A Click event is created when the mouse button is released, not depressed
     Click(MouseClick),
     /// Drag event is created when the mouse was moved over a certain threshold while a button was depressed
@@ -95,11 +95,11 @@ impl MouseDragEvent {
     }
 }
 
-impl SimpleMouseEvent {
+impl MouseEvent {
 
     /// Returns a new copy of the event data relative to the given point
     pub fn relative_to(&self, xy: Point) -> Self {
-        use self::SimpleMouseEvent::*;
+        use self::MouseEvent::*;
 
         match self {
             &Click(mouse_click) => Click(mouse_click.relative_to(xy)),
--- a/src/mouse/events.rs
+++ b/src/mouse/events.rs
@@ -5,6 +5,7 @@
 use input::MouseButton;
 use position::Point;
 use time::SteadyTime;
+use ::vecmath::vec2_sub;
 
 #[cfg(test)]
 use position::Scalar;
@@ -17,7 +18,7 @@ use position::Scalar;
 pub enum MouseEvent {
     /// When the mouse button has been pressed but not yet released or dragged.
     /// This event is created as soon as the button is pressed.
-    Down(MouseButtonDown),
+    Down(ButtonDownEvent),
     /// Indicates that the mouse was clicked. A Click event is created when the mouse button is released, not depressed
     Click(MouseClick),
     /// Drag event is created when the mouse was moved over a certain threshold while a button was depressed
@@ -54,6 +55,16 @@ pub struct MouseDragEvent {
     pub button_released: bool
 }
 
+/// Event that is created when a mouse button if first pressed (but not yet released)
+/// a ButtonDownEvent will always precipitate either a Click or a Drag event.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct ButtonDownEvent {
+    /// The mouse button that was pressed
+    pub mouse_button: MouseButton,
+    /// The position of the mouse when the button was pressed
+    pub position: Point,
+}
+
 /// Holds info on when a mouse button was depressed or released.
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub struct MouseButtonDown {
@@ -77,8 +88,6 @@ impl MouseClick {
 
     /// Returns a new copy of the event data relative to the given point
     pub fn relative_to(&self, xy: Point) -> MouseClick {
-        use ::vecmath::vec2_sub;
-
         MouseClick{
             position: vec2_sub(self.position, xy),
             ..*self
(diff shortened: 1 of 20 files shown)


mitchmindtree:
Sounds interesting! I'll review this shortly :+1:

src/mouse/events.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// Scroll event is created when whenever the scroll wheel is moved
+    Scroll(Scroll),
+}
+
+
+/// Info on a simple mouse click event. This event gets dispatched when a
+/// mouse button goes down then up without moving more than the drag threshold
+/// while the button is depressed.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct MouseClick {
```
mitchmindtree: We are already in the `mouse::event` module, so we could probably change `MouseClick` to just `Click`. This way accessing it via the module would be a little cleaner, i.e. `mouse::event::Click`. We can still re-export it from the lib.rs as `MouseClickEvent`, i.e. `use mouse::event::Click as MouseClickEvent` if we deem it necessary.

src/mouse/events.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// The Point describing the click location
+    pub position: Point
+}
+
+/// Info on a simple mouse drag event. This event gets dispached when a mouse
+/// button is depressed and the mouse is moved a distance greater than the
+/// drag threshold. Holds the start and end positions of the drag, as well as
+/// whether the mouse button is still being depressed.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct MouseDragEvent {
```
mitchmindtree: This could be just `Drag`

src/mouse/events.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// The current time and location of the mouse
+    pub current: MouseButtonDown,
+    /// This will be false if the button is still being held down, or true if the button was released
+    pub button_released: bool
+}
+
+/// Event that is created when a mouse button if first pressed (but not yet released)
+/// a ButtonDownEvent will always precipitate either a Click or a Drag event.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct ButtonDownEvent {
```
mitchmindtree: `Down`

src/mouse/mod.rs
```diff
@@ -0,0 +1,506 @@
...
+
+}
+
+fn distance_between(a: Point, b: Point) -> Scalar {
+    let dx_2 = (a[0] - b[0]).powi(2);
+    let dy_2 = (a[1] - b[1]).powi(2);
+    (dx_2 + dy_2).abs().sqrt()
+}
+
+impl<'a> ::std::iter::Iterator for MouseEventIterator<'a> {
```
mitchmindtree: The `Iterator` trait is in the prelude - no need to use the `::std::iter::` namespace here.

src/widget/drop_down_list.rs
```diff
@@ -280,7 +282,7 @@ impl<'a, F> Widget for DropDownList<'a, F> where
 
                     MenuState::Closed
                 // Otherwise if the mouse was released somewhere else we should close the menu.
-                } else if global_mouse.left.was_just_pressed
+            } else if global_mouse.buttons.get(MouseButton::Left).was_just_pressed
```
mitchmindtree: Seeing as this is the most common case, perhaps there could be a `left` helper method for easier access that doesn't require importing `MouseButton`, i.e. this would allow `global_mouse.buttons.left().was_just_pressed`.

src/widget/text_box.rs
```diff
@@ -340,6 +273,132 @@ impl<'a, F> TextBox<'a, F> {
 
 }
 
+fn get_clicked_elem<C, F>(text: &str, point: Point, args: &UpdateArgs<TextBox<F>, C>) -> Elem
+                                                                        where C: CharacterCache,
+                                                                        F: FnMut(&mut String) {
```
mitchmindtree: This formatting seems unconventional, if you could try and follow the styling of the rest of conrod that would be preferred, i.e.
```rust
fn get_clicked_elem<C, F>(text: &str, point: Point, args: &UpdateArgs<TextBox<F>, C>) -> Elem
    where C: CharacterCache,
          F: FnMut(&mut String)
{
    ...
}
```

mitchmindtree:
Hmmm, so here are a few thoughts that come to mind while reading through this:

1. This proposes moving to a more `Event` oriented handling of user input rather than just tracking the "current state" which was previously the case. I'm very much in favour of this change (see #569) however rather than only making these changes for the `Mouse` related input events, it'd be nice to make this changeover for *all* input events (keyboard as well).

2. Much of the new event code in the mouse modules looks like a second version of the events that we already receive via the `Ui::handle_event` method. Perhaps we could just store those events directly rather than mapping them to a different mouse specific event format, as storing the more general `Input` event would also probably better address the issue I mention in my first point? Would be good to get @bvssvni 's thoughts on this, as he designed the original `Input` enum which IMO might be a better solution to this.

3. Your event enum tree seems a little inverted, i.e. you currently have something along the lines of:
    ```
    MouseEvent
        Scroll
        Down
            Button
        Click
            Button
        Drag
            Button
    ```
whereas this would probably be better described as
    ```
    MouseEvent
        Scroll
        Button
            Down
            Click
            Drag
    ```
However this is probably doesn't matter if we decide to just store the events given via `Ui::handle_input` anyway.

4. I like the idea of automatically handling the capturing/uncapturing of widgets within the `Ui`! I see that you're doing it for capturing keyboard input via a new `widget_under_mouse_captures_keyboard` method, however we might as well make this the case for the mouse also? This might remove the need for a lot of the widgets to manually call `capture_mouse` or `uncapture_mouse`. I'll have to put some more thought into this.

In general, I like many of the ideas in this PR, however I think they could be implemented in a simpler, more general manner by having the `Ui` store a `Vec<InputEvent>`, perhaps in an `InputEvents` wrapper that provides various methods for filtering through all the events in different ways, depending on whether or not the widget that is trying to access them is capturing or not.

bvssvni:
The idea behind the design of `T: GenericEvent` is that you can add custom events and the widgets will receive them. I suggest `Vec<T: GenericEvent>` if that's possible.

psFried:
Excellent feedback, all. I definitely think that having a generic `UpdateAgrs::events()` method makes a lot of sense. I will definitely look into the idea of reusing the existing events. The reason that this initial attempt uses a whole new events module is that there is technically a many-to-one relationship between `GenericEvent`s and `MouseEvents`. The purpose of the `Mouse` was to aggregate one or more `GenericEvent`s into a single `MouseEvent`. This is because there may be more `GenericEvent`s per second than there are Ui  updates per second.

I really think that just having a single definition of all the events that a widget can handle makes a lot of sense, though. I was actually thinking that even things changing focus should just be handled with generic events. 

@mitchmindtree your point #4 is spot on. I think having the Ui handle switching which widgets capture input is important. I don't think that it makes sense to have the it capture the mouse, though. The way i was thinking about capturing is perhaps a little different than how it has been used. Previously, no widget would be given mouse input unless it was capturing. With this change, any widget would just be given the mouse input if the mouse is over that widget. So, the only reason to ever 'capture' the mouse would be in special situations where you want to prevent other widgets from handling mouse input.

I'll take a look into some other approaches this weekend and see what ideas I can come up with.

psFried:
I took some time to look into the code around `GenericEvent` and I think I agree that using those is the right way to go. The general problem of aggregating many low-level mouse events (like button-down, button-released, etc.) into higher-level events (i.e.- double-click) is still one that I think requires a general solution. That is totally solvable, though, by having some sort of `InputsEvents` wrapper that wraps a `Vec<T: GenericEvent>` of events since the last update. We could have a conrod module that interprets a series of low-level events and inserts high-level events into the stream. I actually really like the idea of widgets getting all events (both low- and high-level) from a single source. 

We'll just have to figure out a nice clean api for widgets to filter only the events they care about. This is something that I foresee being rather painful given my current understanding of how `GenericEvent`s work. I'm assuming we'll ultimately have to filter on the `event_id`, but not positive about that. I'll go ahead and close this PR and start on that as soon as I get some time.

PR #665, opened 2016-01-10
Correct handling of KidArea padding in scroll.rs. Re-draw on window resize events.

mitchmindtree:


mitchmindtree:
Published as **0.29.1**.

Issue #666, opened 2016-01-10
Rename graph::algo::`visible_area_of_widget` to `cropped_area_of_widget`.


mitchmindtree:
`visible_area_of_widget` is too general, and could be re-used later for an algo that *actually* finds the total visible area of the widget (i.e. that considers window size, covering, non-blended widgets, etc).

BenMcH:
Should this also change method names such as visible_area_of_widget_maybe_within_depth?

mitchmindtree:
Yep.

BenMcH:
I just used a find and replace and it didnt seem to break, Mitch. Let's see what travis has to say.

mitchmindtree:
Closed by #667.

mitchmindtree:
Hmmm @BenMcH what are your thoughts on this:

I'm now wondering whether or not the function name actually be `uncropped_area_of_widget`?

What the function actually does is return the rectangular area of the widget that is still visible after it has been cropped (at the moment, cropping only occurs when cropping widgets to a scrollable container).

Which do you think makes more sense for this behaviour: `cropped_area_of_widget` or `uncropped_area_of_widget`? Or maybe there's something else that is more suited?

BenMcH:
Hmm, this is an interesting problem @mitchmindtree. I think that for this case, I like cropped_area_of_widget because uncropped_area_of_widget would seem to return the full size of the widget before the crop into a scrollable container.

mitchmindtree:
Oh yes, good point :+1:

PR #667, opened 2016-01-10
Changed all instances of "visible_area_of_widget" to "cropped_area_of_widget" because of issue #666

BenMcH:
I made the change requested in #666 with a simple find and replace. Conrod builds just fine on my machine and runs examples as well.


mitchmindtree:
Looks good, thanks!

As this is a breaking change I'll wait to publish this 'till a little later, along with some other commits that I have on the way.

Issue #664, opened 2016-01-10
Optimisation Potential


mitchmindtree:
I'm using conrod for a pretty intensive GUI project atm:

![screen shot 2016-01-10 at 2 13 31 pm](https://cloud.githubusercontent.com/assets/4587373/12219730/62bd4f48-b7a4-11e5-9fe2-60b603f381f1.png)

When instantiating a huge numbers of widgets (you can't see in the above picture, but there's about ~1500 in the total scrollable timeline) in an animated GUI (the playhead is always moving while audio is playing back), I begin running into some performance issues as the GUI thread begins to choke ~100%cpu.

After doing some profiling, it appears pretty clear that the biggest bottleneck is the drawing of each primitive widget in the backends/graphics module. You can see the GL calls taking almost 50% of the main thread here:

![screen shot 2016-01-10 at 2 54 43 pm](https://cloud.githubusercontent.com/assets/4587373/12219888/41dd2ef0-b7aa-11e5-89d3-85311d50738b.png)

This was anticipated and is a current issue in conrod that has been discussed pretty extensively in IRC that is related to the current way we draw graphics in piston-graphics. There are a lot of options for addressing this:
- @bvssvni has mentioned the potential for a [retained API](https://github.com/PistonDevelopers/piston/issues/1024).
- tomaka has mentioned that we should look into using a texture atlas for more efficient rendering.

However I think there are some more immediate changes we can make for much shorter term gains:
- Track widget visibility in more detail, only drawing widgets that have some area visible within the window. This may vastly reduce the number of GL calls, improving many cases:
  - Do a basic check in the draw function that ensures `draw` is only called on widgets whose bounding `Rect`s are currently within the bounds of the window.
  - Do not draw widgets hidden outside of their scrolling parent's cropped kid_area.
  - Don't draw widget's that are entirely hidden by widgets at a shallower depth. This may be trickier, as we would need to check the alpha channel of the covering widgets.

There is also room for significant gains in conrod itself:

![screen shot 2016-01-10 at 3 12 20 pm](https://cloud.githubusercontent.com/assets/4587373/12219936/998a785e-b7ac-11e5-8bd0-01dd01524be8.png)
1. Setting edges in the graph:
   
   ![screen shot 2016-01-10 at 2 24 49 pm](https://cloud.githubusercontent.com/assets/4587373/12219797/f3ee0894-b7a5-11e5-93a4-28cc9293410f.png)
   
[...]

PR #668, opened 2016-01-10
Draw optimisation. Range::overlap bug fix.

mitchmindtree:
This removes a several bottlenecks I was running into in a personal project, mentioned in #664:

1. Only draw widgets if they're within the window.
2. Only draw widgets if they're not cropped out.
3. Update to latest daggy version with optimised `add_edge` method.

This PR also bumps the version in order to publish the recent graph::algo::visible_area_of_widget -> cropped_area_of_widget function name change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.29.1"
+version = "0.30.0"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Published as **0.30.0**.

bvssvni:
Did the performance improve after this change?

mitchmindtree:
It improved significantly in my timeline project :) will do some more profiling soon to give specifics.

Note that this probably won't have much impact on the examples at all, as in those examples almost all widgets are always visible. In my timeline project, only ~10% of widgets are visible at any point in time, so was a big speedup.

PR #673, opened 2016-01-12
Add `widget_style!` macro for vastly simplifying widget Style struct definitions and implementations.

mitchmindtree:
**WIP - Do not merge yet.**

Pretty stoked with this one! We're able to remove a huge amount of style-related code in favour of a macro that is, imo, much more readable than the verbose Style impls that were everywhere before.

Todo:
- [x] Review remaining widgets to see if either their styling can be refactored to utilise this, or if the `widget_style!` macro can be extended to handle their unique cases.
- [x] Add thorough documenation to the `widget_style!` macro explaining how everything works and why it's necessary.

PR #674, opened 2016-01-12
Add a macro for simplifying builder method implementations. Update widgets to take advantage of it.

mitchmindtree:
Another macro effort to cut down on boilerplate.

So far I've been trying to be quite conservative with the use of macros in conrod in order to retain easy readability and immediate familiarity to new users. Often, deciphering complex macros can feel like learning a new language just to understand some src.

However, I think the wins here are undeniably worth it, and imo the src still seems as readable - perhaps even more concise now that a lot of unnecessary noise has been removed. I'm definitely keen to get other opinions on this though, as I did write the thing :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.30.0"
+version = "0.31.0"
 authors = [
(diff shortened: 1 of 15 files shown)


mitchmindtree:
Published as **0.31.0** along with the `widget_style!` stuff in #673.

PR #675, opened 2016-01-13
rustfmt'd the src directory

sezna:
Requested in #600, I ran rustfmt on the src directory and then double checked the result, removing extra whitespace left by rustfmt.

A few lines exceeded the maximum line length of rustfmt - I fixed some of them, but not all of them because I am not sure if having lines of that length is a style guideline or just a limitation of rustfmt. The lines that caused that error are here:

line 399 in color.rs.
lines 523, 535, 547, 558, 677, 751, 752, 753, 755, 763 in graph/mod.rs.
lines 185 and 1049 in widget/mod.rs.
lines 141 and 161 in widget/style.rs.

PR #676, opened 2016-01-14
changed if statements to min function

sezna:
Changed if statements to the min function for simpler syntax. It was discussed in #675. I will leave this up for a bit, just in case, and then merge, since it is such a small change.


mitchmindtree:
This is incorrect, it should be:

```rust
let l = max(a_l, b_l);
let r = min(a_r, b_r);
let b = max(a_b, b_b);
let t = min(a_t, b_t);
```

I'm going to revert this commit for now.

PR #680, opened 2016-01-18
Add a few simple builder methods to improve ergonomics and retain symmetry of builder pattern in many cases.

### LINKED ISSUE #679: Add methods to `Widget` to assist in conditionally calling builder methods.
Right now, calling certain builder methods on widgets conditionally can be ugly.

For example, say we have a button:

```rust
Button::new()
    .wh(wh)
    .color(PURPLE)
    .react(|| do_stuff())
    .set(IDX, &mut ui);
```

Let's say we only want to set the dimensions (`.wh(wh)`)under some condition, otherwise allowing conrod to infer them from the context. To do this, we must break out of the builder pattern and create a local binding like so:

```rust
let mut button = Button::new();
if some_condition {
    button = button.wh(wh);
}
button
    .color(PURPLE)
    .react(|| do_stuff())
    .set(IDX, &mut ui);
```
which is quite an eye sore.

It would be useful if conrod provided helper methods that made this nicer. Perhaps something like `and_if` as follows:

```rust
Button::new()
    .and_if(some_condition, |button| button.wh(wh))
    .color(PURPLE)
    .react(|| do_stuff())
    .set(IDX, &mut ui);
```

Here, `and_if` only calls the given builder function and returns the resulting `Button` if `some_condition` is `true`. Otherwise, it just returns the original `Button`. This means we don't have to break out of the builder pattern, keeping the code symmetric and readable.

We could also solve the problem of optional building similarly, perhaps with an `and_then` method:

```rust
Button::new()
    .and_then(maybe_wh, |button, wh| button.wh(wh))
    .color(PURPLE)
    .react(|| do_stuff())
    .set(IDX, &mut ui);
```
which would save breaking out of the builder pattern to do `if let Some(wh) = maybe_wh { ... }`.

These methods would probably best exist in some `Builder` trait that was impl'd for all `T` (perhaps even in std?), however I think the most user-friendly option for conrod is to provide these methods via the `Widget` trait to avoid importing yet another trait.

mitchmindtree:
Closes #679

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.31.0"
+version = "0.31.1"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -280,8 +280,8 @@ fn set_widgets(ui: &mut Ui, app: &mut DemoApp) {
         let label = format!("{:.*}", 2, value);
 
         // Slider widget examples. slider(value, min, max)
-        if i == 0 { Slider::new(value, 0.0, 1.0).down(25.0) }
-        else      { Slider::new(value, 0.0, 1.0).right(20.0) }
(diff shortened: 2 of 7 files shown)


mitchmindtree:
Uploaded as **0.31.1**.

PR #688, opened 2016-02-06
Fix bug where Canvas::flow_left would actually flow downwards. 

### LINKED ISSUE #687: Canvas::flow_left flows down instead of left
Looks like `Canvas::flow_left` [was mistyped](https://github.com/PistonDevelopers/conrod/blob/6734b030d57a55bf49b84bc09965d7dba323947e/src/widget/canvas.rs#L178-L181) to flow down instead of left (`Direction::Y(Backwards)` instead of `Direction::X(Backwards)`), making it act exactly like `Canvas::flow_down`.

mitchmindtree:
Closes #687

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "conrod"
-version = "0.31.2"
+version = "0.31.3"
 authors = [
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Fix published as **0.31.3**.

Issue #171, opened 2014-11-19
Images on buttons


bvssvni:
See reddit comment http://www.reddit.com/r/rust_gamedev/comments/2mlvxa/can_conrod_buttons_have_images/

xilec:
In my opinion, it would be awesome realize something like [WPF Content Model](https://msdn.microsoft.com/en-us/library/bb613548(v=vs.110).aspx). I understand that it is too complicated solutions for topic task. But I saw attempts to realize [Layouts](https://msdn.microsoft.com/en-us/library/ms745058(v=vs.110).aspx) in conrod and it would be greate using layouts to evade implementation of controls by hand which is just a combination of some basic controls (like Label, Button and Image for topic case).

mitchmindtree:
Just thought I'd mention that #696 adds a primitive `Image` widget which should make it easy to add image support (or variants) for various widgets.

pierrechevalier83:
I have been trying to tackle some rudimentary implementation of this feature:
https://github.com/PistonDevelopers/conrod/pull/783

I don't understand why if I cover a part of the button with an image, though that part seems not to be clickable anymore. It is obvious when running the all_widgets example from my pull request and clicking the button that has a texture.

Does this behaviour make sense to anyone following this? Any tip?

pierrechevalier83:
For a quick update: I figured out what was missing. The image widget was capturing the mouse events that should have been captured by the button widget.

I know have the basic functionality working: a button with a texture. It doesn't handle positioning of the image within the button with respect to a label, though but it gives a starting point.

I also removed the part that modifies the button in the all_widgets example and wrote a dedicated button example to showcase the feature (heavily inspired from all_widgets).

This included modifying the git history, so I closed pull request #783 and opened #785.

I would welcome feedback, be it code review or suggestions for tackling more complex workflows (Like CompositeButton described in earlier comment)

pierrechevalier83:
Merged pull request for simple fix (add image method that takes image index)

PR #691, opened 2016-02-12
Update to latest changes in graphics and piston_window crates. Removes some unused dev-dependencies.

mitchmindtree:
**Don't merge yet.**

Since updating, all of the colours in all of the examples have become much lighter:

Before
screen shot 2016-02-13 at 1 42 58 am

After
screen shot 2016-02-13 at 1 41 18 am

It's possible that there might be some other change we have to update to in conrod, but perhaps also possible that there may have been a bug introduced to the graphics or gfx_graphics crates during the update to the latest gfx. Will look into this soon as this should be fixed before merging this PR. Ping @bvssvni

PR #696, opened 2016-02-16
Add new `Image` primitive widget. Add `Backend` trait. Change Ui::set_widgets function param from `Ui` to `UiCell`.

### LINKED ISSUE #645: It's too easy for a user to forget to call `Ui::set_widgets` when instantiating their widgets.
At the moment, the `Ui::set_widgets` method takes a function that takes a `&mut Ui` as its argument so that widgets can use this in their `.set(ID, ui)` method calls.

What we should really do, is change that argument to a special wrapper type around a `&mut Ui` (similar to `UiCell`, but more specific to this case) and disallow passing a `&mut Ui` directly to the `.set` methods. This would ensure a user *has* to call `.set_widgets` in order to be able to actually set their widgets.

We should ensure that this wrapper type provides all the same convenience methods as the `Ui` does (apart from `set_widgets` of course).

We should fix this asap as it has already caused issues (see #643).

### LINKED ISSUE #647: `Texture`/`Image` primitive widget.
This primitive could perhaps then be used as the basis for an `Image` widget? Will draft this up further soon.

mitchmindtree:
**WIP - do not merge.**

I think this is most of the way there, however the generic texture associated with the `Graphics` backend is proving to be quite tricky to resolve between `Image` instantiation and `Ui::draw`. Will continue type juggling tomorrow and see if I can fix it :+1:

Closes #647
Closes #645

Relevant to #171 

Edit:

Main breaking changes to look out for in this PR are:

- `Ui::set_widgets` function argument now takes a `UiCell` instead of a `Ui`. To make this adaptation work, the `UiCell` widget-specific methods have been removed now, so methods like `UiCell::input` now require the index of a the widget you want to get input for (see the changes to the widgets in this PR for an example of this change).
- The `Ui` now takes a `Backend` type parameter, and as a result this type param has propagated through to some other places in the conrod src. See the changes to the `examples` on how to update to this change.


mitchmindtree:
Ok, this is going to either require adding a second type param to `Ui` for `Texture` types (i.e. `Ui<C, T>`) or adding a new `Backend` trait with all the necessary associated types used by conrod.

I think I'm going to opt for the `Backend` trait, as the type-param noise reduction should be worth the implementation of a `Backend` trait for each backend (which just involves specifying the `Graphics` and `CharacterCache` types).

mitchmindtree:
I also think I'm going to give the `Widget` trait a type param for the `Backend` (`Widget<B>`). This will allow widget designers to allow for some widgets to take advantage of backend specific details, and have other widgets remain generic over any backend. It will also help to resolve `CharacterCache` and `Texture` types by using the same `Backend` type param for the widget implementation as we do for the `Ui` passed to the widget's update.

mitchmindtree:
Ok, the `Backend` trait seems to be a nice addition for consolidating the set of backend types used by conrod, however it is difficult to add it as a type param to the `Widget` trait in its current state.

The reason for this is that, although we anticipate that the `Widget`'s generic type `B` would be inferred by passing the `&mut Ui<B>` to the `Widget::set` method, this is not the case. This is because `Widget::set` actually takes some generic type `U` which can return a `&mut` to some `Ui<B>` - rustc is unable to infer the `B` through the indirection of this generic `U` type and `UiRefMut` trait bound.

I think the right solution is to change the `Widget::set` generic type to use the solid type `UiCell<B>`. I'll explain below.

At the moment, there are only two types that implement the private `UiRefMut` trait; `Ui<B>` and `UiCell<B>`. `UiCell<B>` was originally designed as a wrapper around `Ui<B>` that restricts the API slightly so that widget's can access the `Ui` within their `update` methods without mutating it outside of the intended manner. Since the `UiCell<B>` was implemented, we also introduced the `Ui::set_widgets` method, which assists conrod's event and widget management by requiring that the user instantiate widgets within a specific scope managed by the `Ui` (the function passed to the `set_widgets` method). However, although the change was implemented and is usable, the API has not yet been made type-safe, which has already caused unexpected behaviour for users. See #645. It turns out that, in the `set_widgets` function, we want to restrict the user from interacting with the `Ui` in the same way that we want to restrict `Widget`s from interacting with it in their `update` methods. Thus, changing the `Ui::set_widgets` method's function argument from `Ui<B>` to `UiCell<B>`, and changing `Widget::set` to take *only* a `UiCell<B>` may be a nice solution for killing two birds with one stone.

mitchmindtree:
So, when changing the `Widget` trait to include `B: Backend` type parameter, I also had to change the `Positionable` and `Sizeable` traits to include a `B: Backend` type parameter in order to constrain the generic `B` type when doing a blanket implementation of both `Positionable` and `Sizeable` for `Widget<B>`.

I believe these generic implementations are part of the reason why, despite making the `Widget::set` related changes I mentioned in my previous comment, rustc is ***still*** unable to infer the type of `B` when all three of `Widget`, `Positionable` and `Sizeable` methods are used within a widget builder pattern. I'm going to continue experimenting to see if there is another way around this.

mitchmindtree:
Woohoooo

screen shot 2016-02-22 at 5 44 32 pm

mitchmindtree:
OK, I think this is pretty much done.

Just going to wait till #691 gets merged, then will rebase and be ready to go :+1:

mitchmindtree:
I've removed the "feature" approach to `Backend` implementations in favour of providing a blanket implementation for all `(T, C)` where `T` is a texture, `C` is a CharacterCache and both satisfy the necessary bounds (as discussed in #699). Seems to work nicely in all of the examples, and shouldn't conflict with a user's own unique `Backend` implementation if they should choose to do so :+1:

mitchmindtree:
Published as **0.32.0**.

Issue #692, opened 2016-02-14
Line is not drawn


andresv:
Mac OS X 10.11.3
Conrod 0.31.3

Here is following snippet:

```Rust
#[macro_use] extern crate conrod;
extern crate find_folder;
extern crate piston_window;

use conrod::{
    Labelable,
    Positionable,
    Sizeable,
    Colorable,
    Frameable,
    Rectangle,
    FramedRectangle,
    FramedRectangleStyle,
    Place,
    Theme,
    Text,
    TextBox,
    Line,
    LineStyle,
    Circle,
    ShapeStyle,
    Ui,
    Widget
};

use conrod::color::{rgb, rgba};

use piston_window::{EventLoop, Glyphs, PistonWindow, UpdateEvent, WindowSettings};

fn main() {

    // Construct the window.
    let window: PistonWindow = WindowSettings::new("Garage Inc", [800, 480])
        .exit_on_esc(true).build().unwrap();

    // construct our `Ui`.
    let mut ui = {
        let assets = find_folder::Search::KidsThenParents(3, 5)
            .for_folder("assets").unwrap();
        let font_path = assets.join("fonts/NotoSans/NotoSans-Regular.ttf");
        let theme = Theme::default();
        let glyph_cache = Glyphs::new(&font_path, window.factory.borrow().clone());
        Ui::new(glyph_cache.unwrap(), theme)
    };

    let mut f1_state = false;

    // Poll events from the window.
    for event in window.ups(60) {
        ui.handle_event(&event);
        event.update(|_| ui.set_widgets(|ui| {
            // Create a background canvas upon which we'll place the button.
            // pad 5px from each side
            conrod::Canvas::new().pad(5.0).set(CANVAS, ui);

            // draw title
            Text::new("Garage Inc")
                .mid_top_of(CANVAS)
                .font_size(14)
                .color(rgb(0.2, 0.35, 0.45).plain_contrast())
                .set(TITLE, ui);

            // draw current time
            Text::new("22:01")
                .top_right_of(CANVAS)
                .font_size(16)
                .color(rgb(0.2, 0.35, 0.45).plain_contrast())
                .set(CLOCK, ui);

            // draw relay states
            // for i in 0..8 {
            //     Rectangle::outline([45.0, 45.0])
            //         .align_left_of(CANVAS)
            //         .set(RELAY + i, ui);
            //
            //     Text::new(&format!("Relay {:}", i)[..])
[...]

PR #701, opened 2016-02-23
Fix bug where the Ui would incorrectly infer the parent as the previous widget rather than the current parent.

mitchmindtree:
This *might* be related to #692, but I haven't looked close enough to be sure.


mitchmindtree:
Stable passed, and nightly errors have been addressed in #691, going to merge.

Issue #342, opened 2015-03-28
TextBox wishlist


dcampbell24:
The textbox widget is currently very simple and would be a lot more useful if it supported a set of features that would allow it to be used in a basic chat server. Major features that would be nice to have include:
- multi-line
- read only
- highlight / copy and paste
- Input Method Editor support

I am guessing adding support for the first two should be fairly straight-forward, but the last two may not be. The read only feature is related to #341 since it means updating the text box programatically is an error.

mitchmindtree:
Related #71, #62

mitchmindtree:
Agreed on this, TextBox is certainly a WIP and could use some love :smile: 

Highlight copy/paste would be a great next feature (once the bugs are addressed heh)! As for multi-line functionality, perhaps we could separate that into a `TextArea` widget? I remember thinking this might be a good idea, but I've forgotten why.

Anyway, I'm planning on starting a synth editor repo soon that will have me diving back into conrod, so I should be able to help out soon :+1:

davekong:
OK, great! I'm working on a networked Hearts game and building out the client GUI right now. I am going to spend a little putting together the table display, then come back to these issues to get chat working.

davekong:
Related #205

LaylConway:
When I get to adding copypaste functionality to phosphorus, I'll make the cross platform functions for interaction with the clipboard available as a crate so it can be used in conrod.

bvssvni:
@LaylConway That would be useful!

jdm:
With regards to the clipboard, please see https://github.com/aweinstock314/rust-clipboard/ .

mitchmindtree:
@jdm thanks, looks good!

psFried:
I'd like to add another feature to this discussion. What about the ability to tab between TextBoxes? 

@mitchmindtree to your earlier point about having a separate TextArea widget, I think this is one of the differences. A TextBox would be suited for quickly entering small amounts of text into many small TextBoxes, so you would want to be able to quickly tab between them. A TextArea would be more like a mini text editor. IMO, you wouldn't want tab/enter to automatically move you to the next input field when you're in a TextArea, you'd want it to insert a literal tab/newline character.

andresv:
What about setting text color, there are `color` and `frame_color` : http://docs.piston.rs/conrod/conrod/struct.TextBox.html, but nothing for `text_color`.
[...]

Issue #700, opened 2016-02-23
The New Event-Based User Input System - TODO and bug-fixes


mitchmindtree:
This is a follow up of changes that are yet to be made following the implementation of conrod's new event-based user input system in #684 as discussed in #670.

#### Widgets that still need to be changed over to the new system:

- [x] Button
- [x] DropDownList
- [ ] EnvelopeEditor
- [ ] NumberDialer
- [ ] Slider
- [ ] Tabs (maybe?)
- [ ] TextBox
- [ ] TitleBar
- [ ] Toggle
- [ ] XYPad
- [ ] CircularButton (from custom_widget.rs example)

#### Other related todo:

- [ ] Review the `Widget::drag_area` method and related types and see if these can be removed in favour of using the new `Drag` events.
- [ ] Fix bug where widgets still receive `Clicked` events when covered by another widget (should be using `Graph::pick_widget` method).

mitchmindtree:
@psFried I'm currently working away on the above and I noticed that we're creating drag events upon raw release events - I'm thinking of refactoring this to only generate drag events on raw move events as this feels a little more intuitive to me, but I'm curious to hear what your thoughts are on this first?

mitchmindtree:
Ahh, I just noticed the `in_progress` field in the `MouseDrag` struct.

How about rather than combining state into the one `MouseDrag` event, we add `DragStart` and `DragEnd` events?

Also, I'm thinking of changing `MouseDrag` -> `Drag` and `MouseClick` -> `Click`, thoughts?

mitchmindtree:
Going to close this issue in favour of tracking progress at #703

PR #703, opened 2016-03-02
[Breaking Change] Refactor Event system and update Widgets accordingly. Add new multi-line TextEdit widget. Remove multiple crate deps, and more.

### LINKED ISSUE #602: When scrolling a widget, have the remainder scroll amount applied to the parent scrolling widget if there is one.
This will allow for the behaviour where once a user scrolls to the bottom of a widget, the scroll is then applied to the containing scrollable widget if there is one.

### LINKED ISSUE #639: Add `Widget::crop_kids` method that crops all `Depth` child widgets of a widget to its area.
We could use the cropping code that already exists for scrolling and then have `.scrolling` just call `.crop_kids` internally.

### LINKED ISSUE #642: Make it possible to send "scroll actions" to the Ui (without using the scrollbar).

### LINKED ISSUE #708: Track the `Instant` at which each `UiEvent` is created once Rust 1.8 lands
Looks like the `Instant` API in std will stabilise in the next stable rust release. I was going to do this using `PreciseTime` from the `time` crate in #703, but it looks like we might as well wait a few days and use the `std` type instead which should make our use of the `time` crate redundant.

This will be useful for creating time-oriented events such as `DoubleClick`

### LINKED ISSUE #711: Encapsulate scrolling logic in a single `Scroll` (i.e. scrollable container) widget
Right now, scrolling logic is baked into conrod itself as a special case. This is slightly unfortunate, as it would be nice to be able to treat scrollbars and scrollable containers just as any other widget is treated. Ideally, we'd have a single `Scroll` widget, which is able to crop and scroll its children widgets. This would allow us to remove the need for tracking `Scrollbar`s specially within the graph (and in turn allow us to remove the `depth_order::Visitable` type). We could also remove all the scroll logic that is currently nested within the `widget::set_widget` function.

There are a few reasons why scroll logic is baked into conrod itself right now:

1. **Scissor / Cropping**: right now there is no way for a widget to "crop" its children directly - there is a special case for scrollable widgets. It should be possible for us to make it so that widgets can simply call a method like `.crop_kids`, so this point shouldn't hold us back.

2. **Applying `xy` offset to child widgets**: there is currently no way for a parent to offset its children widgets (or for a widget to affect the co-ordinates of other widget whatsoever). Normally I think it can be a little dangerous to allow widgets to "mutate" other widgets, however I think it should be reasonable for us to add a method which would allow for a parent to offset its children widgets. I.e. `.offset_kids`.

3. **Calculating the bounding box of all children in order to calculate the scrollable range**:  this should now be possible to do from within a widget, as these algorithms are exposed publicly from the `graph::algo` module.

4. **"Draining" scroll events**: This currently isn't implemented as it stands, but it is unclear how we would implement this in a custom widget. Basically; remaining scroll that is not used by a child scrollable container should roll over to the parent scrollable widget. Perhaps we could provide a method `ui.drain_scroll(xy) -> Option<Point>` which would allow for a widget to "drain" a certain amount of scroll from the `Ui`'s stored `Scroll` events? Will have to think about this some more.

I think I'll avoid updating the scroll logic in #703 in favour of including this widget in a follow up PR.

### LINKED ISSUE #724: warning: broken MIR (occurs when compiling any of the examples on rustc stable 1.8)
I don't know if this matters, but thought I'd let you know. The all_widgets example seemed to run okay except that I couldn't change the numbers for frame height or width in the GUI: I tried using the mouse wheel, clicking above, below, and on the number, typing a number and none had any effect.

The same warning also occurs with canvas, primitives, and text. Incidentally, the text example is very slow to resize (I'm on a 3+GHz i7).

examples/all_widgets.rs:148:39: 148:44 warning: broken MIR (var2 = conrod::ui::Ui<B>::new(tmp19, tmp21) -> [return: bb44, unwind: bb43]): call dest mismatch (conrod::ui::Ui<(gfx_texture::Texture<gfx_device_gl::Resources>, gfx_graphics::glyph::GlyphCache<gfx_device_gl::Resources, gfx_device_gl::factory::Factory>)> <- conrod::ui::Ui<(<gfx_graphics::back_end::GfxGraphics<'static, gfx_device_gl::Resources, gfx_device_gl::command::CommandBuffer> as graphics::graphics::Graphics>::Texture, gfx_graphics::glyph::GlyphCache<gfx_device_gl::Resources, gfx_device_gl::factory::Factory>)>): Sorts(ExpectedFound { expected: gfx_texture::Texture<gfx_device_gl::Resources>, found: <gfx_graphics::back_end::GfxGraphics<'static, gfx_device_gl::Resources, gfx_device_gl::command::CommandBuffer> as graphics::graphics::Graphics>::Texture })
examples/all_widgets.rs:148         Ui::new(glyph_cache.unwrap(), theme)
                                                                  ^~~~~
     Running `target/release/examples/all_widgets`

mitchmindtree:
~~**WIP - do not merge yet.**~~

These changes are a follow up on the implementation of conrod's new event-based user input system in #684 as discussed in #670.

#### Related todo:

- [x] Have `GlobalInput` take `drag_threshold` as a push_event argument rather than storing it to ensure the value is always up-to-date.
- [x] Track `widget_under_mouse` in `InputState` on `mouse_move`, `resize` and `scroll` so that we can use it to track widget-specific `MouseClick` etc events properly.
- [x] Consider inlining `InputState::update` and event interpretation within `GlobalInput::push_event` as they're only used once in a very specific area (`Ui::handle_event`).
- [x] Provide a `backend::ToEvent` trait which allows users to pass custom events to `Ui::handle_event`, as long as they implement the `ToEven` trait.
- [x] Keep track of the widget that was under the mouse when each button was pressed by storing a `Option<widget::Index>` for `ButtonPosition::Down`.
- [x] ~~`WidgetInput::all_events` currently takes `global_input.start` - shouldn't it take `global_input.current`?~~ Ahh, I see that `InputState::update` was being called as each event was emitted. As long as we track `widget_under_mouse` during handle_event and within each `UiEvent` where necessary, I think we can remove the incremental updating of the `InputState` in the `WidgetInput` event iterator.
- [x] `MouseClick` should track the widget that was directly under it at the time so that it can be delivered to the correct widget (rather than whatever happens to be under the mouse at `current` when `Ui::set_widgets` is called).
- [x] Ensure that all `Input` types are exposed from `backend` module (so that users don't need to import piston-input crate).
- [x] Fix bug where widgets still receive `Clicked` events when covered by another widget (should be using `Graph::pick_widget` method).
- [x] Change module structure from:
```
events/
    global_input
    input_provider
    input_state
    widget_input
    ui_event
```
to
```
event
input/
    global
    provider
    state
    widget
```
- [x] Review behaviour of `widget_input::should_provide_event`. Edit: Removing this along with an overhaul of the input::widget module.
- [x] Review the `Widget::drag_area` method and related types and see if these can be removed in favour of using the new `Drag` events.
- [x] ~~Record the precise time each event was received and store it with each event (see #708).~~ Just tracking the time of the last `Click` event now in order to generate `DoubleClick`s, will hold off on #708 until it is necessary.
- [x] Remove `Ui::Raw` and `Widget::Raw` in favour of adding more event variants that correspond with each of the raw event variants. This should simplify the event API for new conrod users.
- [x] Remove all manual capturing and un-capturing of the mouse - this should now be handled entirely within `handle_event`.
- [x] Remove old `UserInput` type and related data in `Ui`.
- [x] Move text-layout logic from glyph_cache module into a new text module at the top layer. Add a way to produce iterator's yielding `Rect`s for both line and character position:
    - `pub mod char`:
        - [x] widths -> Widths
        - [x] xs -> Xs
    - `pub mod line`:
        - [x] struct Info { start, break, width }
        - [x] struct IndexPair { start, end }
        - [x] enum Break { Newline(end, next_start), Wrap(end, next_start) }
        - [x] struct Infos { .. } // Iterator yielding `Info`s
        - [x] fn rects(infos, bounding_rect, x_align, y_align, max_width, line_spacing) -> Rects
        - [x] fn char_rects(Rects, text, cache, font_size) -> CharRects
    `pub mod cursor`:
        - [x] fn selection_rects<I>(lines_with_rects: I, text, cache, font_size, start, end) -> SelectionRects
- [x] Create `Scroll` events for scrollable widgets on `MouseScroll`. If not over any scrollable widgets, emit `Scroll` events to the widget capturing the mouse if there is one.
- [x] Add `Ui::scroll_widget` and `UiCell::scroll_widget` methods that allow for the user (or other widgets) to emit `Scroll` events to other widgets.
- [x] Apply `Scroll` events to scrollable widgets during the `widget::set_widget` function.
- [x] Remove all built-in `Scrollbar` logic into a new `Scrollbar` widget that emits `Scroll` events to its associated widget via the `UiCell::scroll_widget` method.

Decided to leave the following three points for now in favour of doing #711 instead.
- ~~Drag scrollbars when `Drag` events are produced and produce a `Scroll` event.~~
- ~~Apply `Scroll` events as they are produced (see #709. This will need to be applied within this PR).~~ 
- ~~Track `graph::Visitable` instead of `widget::Index` under pressed mouse buttons (in order to differentiate scrollbars from widgets).~~

#### Events (inspired by [DOM events](https://en.wikipedia.org/wiki/DOM_events)):

- [x] Click
- [x] Drag
- [x] Scroll
- [x] DoubleClick

*Decided to leave the following events for a future PR or until they're needed.*
- [ ] MouseOver (*widget only*) // Mouse moves over the widget ***or*** one of its children
- [ ] MouseOut (*widget only*)
- [ ] MouseEnter // Mouse moves over a widget
- [ ] MouseLeave
- [ ] DragStart
- [ ] DragEnd
- [ ] DragEnter
- [ ] DragLeave
- [ ] DragCancel // Cancelled via pressing `Esc` or some other means.

#### Widgets that still need to be changed over to the new system:

- [x] Button
- [x] DropDownList
- [x] EnvelopeEditor
- [x] NumberDialer
- [x] Scrollbar *(new)*
- [x] Slider
- [x] Tabs (maybe?)
- [x] TextBox
- [x] TextEdit *(new)*
- [x] TitleBar
- [x] Toggle
- [x] XYPad
- [x] CircularButton (from custom_widget.rs example)

Closes #708.
Closes #724.
Closes #711. *Scroll logic is still built in, however the Scrollbar was moved into its own Widget.*
Closes #642.
Closes #639.
Closes #602.
Addresses some parts of #342.


psFried:
@mitchmindtree Just got back and saw your comment on #700. Yeah, the purpose of creating drag events on mouse release was so widgets could differentiate between in-progress and completed drags. I certainly don't mind the idea of separate event for `Drag` and `DragEnd` or something like that. 

On the names, `MouseClick` to `Click`, etc.: I went back and forth on that. The context that I'm thinking about is how would we want this to work if a user uses a controller or joystick to click. Or, if there's programmatically generated click events, or touch events on a touchscreen. I don't know how those things might be provided as raw input events. I originally decided to go with `MouseClick` over `Click` because the raw input events are also mouse specific. Thinking about it now, though, I'm thinking that `Click` is really better because it could be caused by things other than a mouse.

mitchmindtree:
Going to be away for the weekend, but will get back to this on Monday :+1:

mitchmindtree:
I've been working on improving a DSP API over the past week, but will hopefully get back to this within the next couple days.

bvssvni:
Updated to latest Gfx. Here is a dump from Eco:

```
    "dependencies": {
      "pistoncore-input": {
        "bump": {
          "old": "0.8.0",
          "new": "0.9.0"
        }
      },
      "piston2d-graphics": {
        "bump": {
          "old": "0.15.0",
          "new": "0.16.0"
        }
      }
    },
    "dev-dependencies": {
      "piston_window": {
        "bump": {
          "old": "0.37.0",
          "new": "0.39.0"
        }
      },
      "piston": {
        "bump": {
          "old": "0.17.0",
          "new": "0.19.0"
        }
      }
    }
```

mitchmindtree:
Thanks!

mitchmindtree:
Alright, I think this PR is fiiiiinally good to go!

I believe the only thing blocking it now is the ICE occurring when rustc attempts to compile the all_widgets.rs example. I've posted an issue for this [here](https://github.com/rust-lang/rust/issues/34027).

As this PR involved refactoring most of the foundational logic in conrod and in turn updating all of the widgets, it made sense to address some other issues at the same time.

One of the biggest changes is the addition of the `TextEdit` widget. This widget allows for multi-line text editing including selection, deletion, cursor movement (`Up` and `Down` are still not implemented), horizontal alignment, line spacing etc. It is largely the same as the `Text` widget (it actually uses it internally to display the text) however also offers editing capabilities. The `TextBox` widget is now vastly simplified and uses a `TextEdit` widget internally. The `TextEdit` is still missing some features including:
- Add newline on `Return` (not sure whether we should enter insert a `\n`, `\r` or `\r\n` here? Maybe there's a std feature that does this that I'm not aware of).
- `Up` and `Down` cursor navigation.
- The `react` function is never called. We should make a `text_box::Event` enum which enumerates the different kind of events that might be emitted and have this passed to the react function.

It should also be noted that all text handling logic has been moved from the `glyph_cache.rs` module into a unique `text.rs` module at the top level. This module has `char`, `cursor` and `line` modules internally for handling related text logic.

Widgets no longer get `Scrollbar`s automatically when calling `scroll_kids`, `scroll_kids_vertically` or `scroll_kids_horizontally`. Instead, `Scrollbar` has been added as a new widget. See an example of how they may be used at the `Canvas` widget in the all_widgets.rs example. This change is beneficial for a few reasons:
- Allows us to remove scrollbar logic from the core of conrod.
- Allows users to opt-in to their `Scrollbar` use (many widgets for specific platforms might not want to expose a scrollbar at all).
- Allows users to create their own custom scrollbar widgets with unique skins, etc.

Cropping of children widgets is no longer limited to scrollable widgets - any widget may now crop its children widgets by calling the `.crop_kids` method.

Removes `rand`, `time` and `vecmath` crates, leaving us with only 4 dependencies: `daggy`, `num`, `pistoncore-input` and `piston2d-graphics`. We should probably also move towards removing our dependency on the deprecated `num` crate.

All-in-all the new event system has allowed us to remove a significant amount of complexity from most widgets! This PR adds two new widgets (`TextEdit` and `Scrollbar`) as well as an entire module for text handling (`text.rs`) yet still removes more code than it adds. One of the key benefits is that widgets no longer have to track their own `Interaction`s and no longer have to manually capture or uncapture input devices, as this is all handled automatically within the conrod `Ui` as events are received. This helps to cut down on a lot of boilerplate (for an example, check out the diff for the `button.rs` widget module).

@psFried @bvssvni

bvssvni:
I could not resist the temptation and tried modify the "text" example. However, I can't figure out the `.react` closure signature. Could you help me, I just NEED TO SEE IT NOW!! I'm so excited 😄

bvssvni:
I figured it out, but got the compiler bug on the modified "text" example. 😢

mitchmindtree:
@bvssvni hmmm strange that you got the bug on the modified version, but not the regular version! Looking into it now to see if I can recreate this and work out what's going on.

Not sure how to side-step this issue just yet. It doesn't look like anyone's looking into the ICE yet, so I'll probably spend the day looking into a workaround to use in the meantime.

mitchmindtree:
Hmmm I got the same thing. Simply changing the widgets in the `text.rs` example from `Text` to `TextEdit`, changing the `DEMO_TEXT` to a mutable `String` and adding the empty `react` methods to each widget causes it to ICE. Curiously, removing the call to the `.set(*, ui);` method stops the ICE from occurring (just like in the `all_widgets.rs` example), despite it working without issue for widgets in other examples. There seems to be something rustc doesn't like about calling `Widget::set` on the `Text` or `TextEdit` widgets.

bvssvni:
Nice to make some progress on that compiler bug!

mitchmindtree:
Alright, ICE is fixed! Never been so happy to see that green tick heh!

@bvssvni don't get too excited about `TextEdit` just yet! Still needs a bit of work [I've opened an issue here](https://github.com/PistonDevelopers/conrod/issues/729) which has the remaining TODO that came to mind (there's probably more I'm forgetting or haven't thought of). Would love for you to have a play with `text_edit.rs` and add some points.

In the meantime I'll bump the crate version, publish and merge this!

mitchmindtree:
Published as **0.36.0** 🎉

psFried:
I've been away for a while, but just wanted to say that these changes look really good. @mitchmindtree

Issue #717, opened 2016-04-30
Examples Don't Run


Bluebird45:
This on Linux x64, Rust 1.8, Rustup, for all the examples, although "cargo test" passes OK:

> 
> $ cargo run --example all_widgets
> Running `target/debug/examples/all_widgets`
> thread '<main>' panicked at 'Failed to get root window: XError { description: "GLXBadFBConfig", error_code: 178, request_code: 155, minor_code: 34 }', ../src/libcore/result.rs:746
> note: Run with `RUST_BACKTRACE=1` for a backtrace.
> Process didn't exit successfully: `target/debug/examples/all_widgets` (exit code: 101)
>

bvssvni:
You could try switching the window backend:

```rust
extern crate sdl2_window;

use sdl2_window::Sd2lWindow;

let window = PistonWindow<Sdl2Window> = ...;
```

piston_window uses pistoncore-glutin_window by default, but you can change to another, for example pistoncore-sdl2_window if there are problems on your platform.

Bluebird45:
OK, I installed sdl2 and made the configuration changes that you suggest. I also downgraded the line in examples/custom_widget.rs that mentions the OpenGL version, from 3_2 to 2_1.

I also changed computers: both are Dells with onboard graphics controller. I made sure I had the latest available drivers installed from Debian's testing repo. 

"cargo test" passes

but running the example fails:

> $ cargo run --release --example all_widgets
> Running `target/release/examples/all_widgets` thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: "Could not create GL context"', ../src/libcore/result.rs:746
> note: Run with `RUST_BACKTRACE=1` for a backtrace.
> Process didn't exit successfully: `target/release/examples/all_widgets` (exit code: 101)
>

Bluebird45:
Here is the same (SDL2) build running backtrace:

> $ RUST_BACKTRACE=1 cargo run --release --example all_widgets
>      Running `target/release/examples/all_widgets`
> thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: "Could not create GL context"', ../src/libcore/result.rs:746
> stack backtrace:
>    1:     0x7fc1e3f96c00 - sys::backtrace::tracing::imp::write::h3675b4f0ca767761Xcv
>    2:     0x7fc1e3f9938b - panicking::default_handler::_$u7b$$u7b$closure$u7d$$u7d$::closure.44519
[...]

PR #718, opened 2016-05-04
Make it easier to change the OpenGL version in the examples.

Bluebird45:
This is a followup to issue [#717](https://github.com/PistonDevelopers/conrod/issues/717 "#717"), to make it easy for beginners with older hardware to step down the OpenGL version specified in the examples.


bvssvni:
Nice!

PR #734, opened 2016-06-14
Remove `Backend` in favour of rendering a `Primitives` list. Remove `CharacterCache` in favour of integrating RustType directly. Support multiple fonts. Add glutin_glium.rs example.

### LINKED ISSUE #576: Conrod should expose (or at least re-export) everything a user needs to use their own graphics and event backends.

### LINKED ISSUE #725: Can't update Texture of an Image
Since `Image` requires the texture to be wrapped in an `Arc`, there is no way to mutate the original texture. Could conrod implement another way of drawing a texture that allows for mutability?

mitchmindtree:
Step-by-step overview of changes
------------------------------------------

**Remove `CharacterCache` in favour of RustType**

This branch started when attempting to replace the `CharacterCache` with the `RustType` crate in order to improve the performance when rendering text while also removing the `Backend::CharacterCache` associated type. `RustType` tackles the same problem as `CharacterCache` - attempting to provide a backend agnostic way of caching and rendering widgets - however it does so without using any traits and in a way that allows for updating a cache on the GPU with a large number of glyphs at once (rather than storing a texture for every individual glyph).

Edit: Also, conrod now supports using multiple fonts!

**Remove `Texture` associated type.**

After removing the `CharacterCache`, I realised that the `Texture` associated type could also be removed. Instead of having the `Image` widget take a `Texture` directly and storing it within the `Ui`, it should just take a `texture::Id` which could be mapped to some texture when required for rendering.

Edit: See further discussion on the `Image` widget at #725. Will most likely pass on this idea of `texture::Id`s in favour of using the `widget::Index`s associated with instantiated `Image` widgets.

**Remove `Backend` trait**

Without either the `CharacterCache` or `Texture` associated types, the `Backend` trait became unnecessary as a result and was removed.

**Render to list of `render::Primitive`s**

In order to take advantage of the `rusttype::gpu_cache::Cache` properly, I needed to refactor the way the `Ui` is drawn. Instead of taking some piston `Graphics` backend, we instead have `Ui::draw` return a list of `render::Primitives` in order of the depth at which they are drawn, starting at the bottom and ending at the top. The `Primitive` variants are:
- `Rectangle`
- `Polygon`
- `Lines`
- `Text`
- `Image`

This way text could be cached within the `rusttype::gpu_cache::Cache` right before the `Primitive::Text` is yielded - this way a user can easily cache the glyphs in their own texture on the GPU and use the `gpu_cache::Cache::rect_for` method along with the list of `PositionedGlyph`s to draw the glyphs themselves. The `Primitives`-list style of rendering also means that the user can map the `texture::Id`s for `Primitive::Image`s to their own texture types without the `Ui` having to bound the texture type in any way.

By rendering as a list of `Primitives`, this also makes it much easier for other non-piston windows and graphics backends to integrate conrod. They just have to call `Ui::draw` and then render the list of yielded primitives in the order that they are produced. To demonstrate this, I've added a `glutin_glium.rs` example (as of writing this i still need to work out how to most efficiently use glium to render the yielded primitives).

**`backend` features**

Following this PR we can think of conrod as a pipe where `conrod::event::Raw`s are the input and `conrod::render::Primitives` are the output. If the above was all that was changed in this PR, we'd be off-loading a *lot* of new work onto users. Users would be required to work out how to convert events and render `Primitives` themselves as well as handle `scizzor`ing, the new rusttype glyph cache, etc.

In order to keep the API as simple and easy to use as it was prior to this PR, we introduce `backend` features. These are optional dependencies and features that provide functionality specific to certain backends so that users can get started with their backend of choice with ease. I've added a brief description of each feature to the Cargo.toml, so rather than repeat myself I'll paste that here:

```
# Optional dependencies and features
# ----------------------------------
#
# None of the following dependencies are necessary to use conrod, however they may greatly
# simplify the work involved when using conrod with certain window or graphics backends.
#
# `glium`
# Provides functions for rendering the `conrod::render::Primitives` yielded by `Ui::draw`.
# Enables the `conrod::backend::glium` module.
#
# `glutin`
# Provides a function for converting glutin `Event`s to `conrod::event::Raw`s.
# Enables the `conrod::backend::glutin` module.
#
# `piston`
# Provides functions for:
# - Converting piston `GenericEvent` types to `conrod::event::Raw`s.
# - Rendering the `conrod::render::Primitives` yielded by `Ui::draw`.
# Enables the `conrod::backend::piston` module.
#
# `piston_window`
# Provides functions for:
# - Text cache `G2dTexture` construction.
# - Rendering the `conrod::render::Primitives` yielded by `Ui::draw`.
# Enables the `conrod::backend::piston_window` module.
# Note: Use the `piston` feature for `piston_window` event conversions.
```

Todo
------

- [x] Consider adding a sixth `PrimitiveKind` variant called `Other` which yields a `&graph::Container` for each non-primitive widget, so that users can render custom widget types however they wish (i.e. widgets that require rendering via some specific kind of shader, etc).
- [x] Remove the `Primitives::draw` method that takes a closure in favour of making `Primitives` an `Iterator` to improve ergonomics. This will probably require turning it into a kind of state machine. *Edit: It turns out we cannot safely provide this as an `Iterator`, however a `next` method was still implemented to allow for use within `while let Some(prim) = primitives.next()` loops.*
- [x] Work out what's causing the `UnitCountMismatch` error when updating the text texture cache in all examples using `piston_window` that draw text. *Edit: fixed by PistonDevelopers/texture#40 and PistonDevelopers/gfx_texture#88.*
- [x] Add a `piston_window` feature
- [x] Test API where user sets constant text cache texture size. This might be safer than the current case where conrod dynamically sizes the rusttyp GPU cache depending on the window size and the user has to keep track of this and update the actual cache texture themselves.
- [x] Improve `Image` API as per discussion at #725 
- [x] Attempt to simplify examples to their original simplicity and improve the feature gated backend modules by alleviating as much boilerplate in each as possible.

Future work
------------
- Consider Adding an agnostic adaptor that renders the `Primitives` to agnostic vertex, index and command buffers (it seems like this is how [imgui-rs](https://github.com/Gekkio/imgui-rs/blob/master/src/lib.rs#L294) does it's agnostic rendering).
- Render the yielded primitives in the `glutin_glium.rs` example. Look into using instancing (see the glium [instance.rs](https://github.com/tomaka/glium/blob/master/examples/instancing.rs) example and maybe checkout how webrender does this). Note that `Text` is already working nicely!
- Add a `glutin_gfx.rs` example - like the `glutin_glium.rs` example, but demonstrating performant gfx rendering instead. May involve adding a `gfx` feature.

Updating your code
-----------------------

If you use conrod to simply instantiate existing widget, see the diff for the examples to see what changes you'll need to make to your own code.

If you design custom conrod widgets, see the diff for conrod's widgets.

Closes #576.
Closes #725.


bvssvni:
Nice!

An idea to make the glyph rendering easier:

- Add `Graphics::factory_mut` to Piston-Graphics to get a mutable reference to the factory
- Make Piston-Graphics depend on RustType directly, replace the CharacterCache trait and simplify the 2D backends

This could simplify the code in Conrod quite a bit.

I like the idea of rendering to a list of primitives! Have you thought about making them higher level, for example with UI style? A backend could use effects for the theme, or maybe make the widgets look native to the platform.

Now that the `Backend` trait is gone, I suggest using `E: GenericEvent` and implement it for an enum that stores the higher level events with the lower level one (See [input::Event](https://github.com/PistonDevelopers/piston/blob/master/src/input/src/event.rs#L11)). This would allow sending custom events to the widgets, which is a use case `GenericEvent` is designed for.

Perhaps use Piston's window backends to map events?

mitchmindtree:
> Make Piston-Graphics depend on RustType directly, replace the CharacterCache trait and simplify the 2D backends

@bvssvni I think that's a great idea! I'm not familiar enough with the different backends to be sure, but I might have a look soon. It would be great if there was some way to properly take advantage of the `rusttype::gpu_cache::Cache` with piston graphics. At the moment, If I run the `text.rs` example (uses piston_window) and draw every frame, it consistently uses about 70-80% cpu on my machine, whereas if I draw roughly the same amount of text in the new `glutin_glium.rs` example (which uses the GPU cache properly) at the same rate, it maxes at about 5% cpu. I'll setup a proper test soon and ping you so you can have a play yourself.

> Now that the Backend trait is gone, I suggest using `E: GenericEvent`

Thanks for the suggestion! I have thought about this and I'd prefer to keep conrod's event system as an enum of known events for a few reasons:

- The primary reason is that the widget API is immediate, so on the rare occasion that a widget does need some unique event type to be available, it is trivial to pass it directly to the widget quite easily (just as the rest of each widget's unique data is passed the same way). This way custom widgets may handle unique events and conrod doesn't have to know about them or box events, cast them etc.
- I'd like to avoid any more generic bloat (it's a big relief to have just gotten rid of the `Backend` trait). This might not really an issue though if `GenericEvent`s can be stored as trait objects.
- I think matching on an `enum` of known events is a bit simpler and clearer for users - when first coming across `GenericEvent` it took me a while to grok what was going on and how it was used, and importing all of the traits either requires a glob import or tediously adding each import.

I can see how it might be handy for a GUI that uses a more retained-style API, however I think it might be overkill for conrod which already provides a simple way of passing unique data directly to widgets every update.

Just have a few more text bugs to fix then I think this PR should be good to go!

cc @pierrechevalier83

pierrechevalier83:
@mitchmindtree, 
> Just have a few more text bugs to fix then I think this PR should be good to go!
Amazing! I've beent testing this branch a little bit and I have to say I'm pretty impressed. The client code is simpler and textures of images can now be updated :)
I have noticed a few glitches with the rusttype fonts, for instance in the "all_widgets" example, so I'm assuming that's what you're refering to here.

mitchmindtree:
@pierrechevalier83 Glad it's feeling a bit nicer! Yeah I introduced a couple bugs when updating the `src/text.rs` module, I've fixed a couple but there's still some strange behaviour with `TextBox` and `TextEdit` widgets. I have a feeling this might have already existed, but might as well fix it now.

mitchmindtree:
Alright, should be good to go! There's still some decisions to be made and further work to be done (see the "Future Work" section in the top comment), but that can wait for further PRs.

PR #735, opened 2016-06-18
Add a new RangeSlider widget along with an example.

mitchmindtree:
The example uses the range slider to drag the left and right edges of the oval below.

screen shot 2016-06-18 at 2 45 49 pm


mitchmindtree:
Doesn't yet support vertical orientation but will merge for now and add it later when I get time.

PR #736, opened 2016-06-21
Add iterator extensions to make yielding specific kinds of press and release events easier.

mitchmindtree:


mitchmindtree:
Published as **0.36.2**.

PR #738, opened 2016-06-28
Add new `KeyPress(Vec<std::path::PathBuf>)` event to `FileNavigator` along with more styling options.

mitchmindtree:
New styling options include:
- custom width for resize handle
- unique colors for selected, unselected and text.

Also adds an auto-scroll when the column width is resized beyond the width of the FileNavigator, or when a directory is added to the stack that would exceed the FileNavigator width.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: rust
 
 rust:
     - stable
+    - beta
     - nightly
 
 notifications:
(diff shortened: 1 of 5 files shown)


mitchmindtree:
Travis fails on nightly, but succeeds on both stable and beta.

mitchmindtree:
Published as **0.36.4**.

PR #747, opened 2016-07-08
Add a method to Scrollbar for setting the thickness.

mitchmindtree:
This PR also addresses some new warnings emitted since updating to rustc 1.10.


mitchmindtree:
Published as **0.37.1**.

PR #748, opened 2016-07-16
Remove unnecessary inline hints causing bloated stack size (and in turn overflowing the stack) in downstream crates

mitchmindtree:
I've run into several stack overflows using conrod over the past couple of days, only when compiling in release. When instantiating many widgets (i.e. via helper macros), the large amount of inlining lead to a huge stack size that overflowed the designated 8mb stack on my OS X release build. This fixes the problem and hasn't caused any noticeable difference in performance. It seems that in this case, it's best to let LLVM automatically infer what cases are best for inlining. Perhaps it uses some sort of main-thread stack size consideration when deciding upon what to inline?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod"
-version = "0.37.1"
+version = "0.37.2"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Travis seems to be failing due to an ICE on nightly (either [this](https://github.com/rust-lang/rust/issues/34780) or [this](https://github.com/rust-lang/rust/issues/34839)), however is building without issues on stable and beta.

mitchmindtree:
Published as **0.37.2**.

Issue #460, opened 2015-05-27
Widget suggestion: List view


Izzeri:
I think this control is called a ListView or something similar in Windows. Basically a drop down list that is always expanded. Has a fixed size, and becomes scrollable if the entries in it take up more space than its height. Entries can be selected by clicking. Maybe allow selecting multiple entries?

mitchmindtree:
I think we should be able to do this with the `DropDownList` widget - perhaps we could add an `.always_open(bool)` builder method?

HitomiTenshi:
Personally I'd prefer having every widget do its thing. Having an `.always_open(bool)` in the `DropDownList` sounds logical but it doesn't feel right since the widget is supposed to `DropDown` as its name implies. I think it's better to create a `ListView` for that scenario, people will find it much more intuitive.

You know the Unix saying "Do One Thing and Do It Well", I think that should also apply to widgets. :smirk:

cyplo:
+1 for the separate widget

BenMcH:
If that is all that is needed to convert the drop down list to this list view, I'll work on this feature either Sunday night or Monday.

mitchmindtree:
Just an idea - it might be simplest to implement this internally as a single `WidgetMatrix` with 1 column and *n* rows of `Button`s where *n* is the number of items in the list.

BenMcH:
Hmm, I am unsure of what we should do and since I am VERY new to rust, I do not want to mess this up. I might return to this later.

mitchmindtree:
There are now both [List](https://docs.rs/conrod/0.51.1/conrod/widget/list/struct.List.html) and [ListSelect](https://docs.rs/conrod/0.51.1/conrod/widget/list_select/struct.ListSelect.html) widgets that should satisfy this issue. Examples for each can be found in the `examples/` directory.

PR #749, opened 2016-07-21
Implemented simple list view with optional multi selection mode.

Hyperchaotic:
Hiya,
I've created a list widget for conrod and would appreciate some feedback.  I started getting to know Rust a few weeks ago and conrod last week so I'm green. I'm coming from C++ and getting to know this exciting new language.

It's a simple list, based on some of the code in file_navigator. Multi selection mode is optional.  Uses Shift/Alt on Mac or Shift/Ctrl depending on platform support. The input Vec takes a Generic implementing the Display trait instead of Strings.
On keypress/mouse it returns (index of item, String). It would be nice returning a copy of the original T but getting the type information across from ListView::new through State through ListViewEvent and back to the user proved very painful so I went with the index+string representation.

I wasn't sure what to call it, ListView? ListBox? SimpleList, etc. so I went with ListView.

It all compiles without errors/warnings and there is an example app that prints the events in the console, e.g. "Event received SelectEntry((2, "Common Snapping Turtle"))". 

Only problem I have is the event object ListViewEvent I return to the caller, see the commented out "match event" code in the example app. I have tried anything I could think of but can't seem to get that enum into scope.

Thank you!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod"
-version = "0.38.2"
+version = "0.39.0"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"


tafia:
I haven't read the code so I'm not sure it is not the case. 
For a listview I'm expecting a control with multiple columns. If there is only one column then I'd call it a listbox

mitchmindtree:
Woah this is such a coincidence - I spent most of yesterday working on a `List` widget myself! I also had the same plan of generalising the logic shared between the `FileNavigator`'s `DirectoryView`, the `DropDownList` etc.

Firstly, thanks so much for sharing - massively appreciated! I've only had a brief look through the src but it certainly looks like a useful widget :+1:

Before accepting and giving a closer review however, I'd like to share some of my own ideas as I think we can make this a little more generalised. I think @tafia's point is an important one and I think a slightly simpler, more generic solution might be possible to achieve this.

The `List` widget I've been envisioning would imply no bounds on the kind of widgets that may be used as items, and would be more of a helper widget that simply simply takes care of list layout and scrolling (kind of similar to the way the `WidgetMatrix` works). The `List` itself wouldn't know anything about the types of the individual items. It would be more of a primitive widget that could be used to create more complex list widgets (i.e. multi-selection, tables, context menus, etc).

The `List` would only take care of only the following:
- Layout of sequential widgets
- Scrolling of list items
- Provide the option of only instantiating widgets for the currently visible items (i.e. not instantiating any widgets that are outside of the current scroll view) for efficiency. `FileNavigator` currently doesn't do this and as a result gets very slow when navigating directories with many entries.
- Generates the necessary `NodeIndex`s for each item in the list.
- Whether the `Scrollbar` is always visible on the side of the list, or whether it uses an auto-hide style.

I imagine instantiating the `List` would look something like this:

```rust
// A simple multi-selection list example.
let mut list = vec![bool; 10];
```

```rust
// Widget instantiation.
List::new(list.len(), ITEM_HEIGHT)
    .scroll_kids_vertically()
    .item(|item| {

        let toggle = Toggle::new(list[item.i])
            .react(|new_is_selected| list[item.i] = new_is_selected);

        // This would assign the position, size and finally `set` the widget.
        //
        // widget
        //     .w_h(self.w, self.h)
        //     .and(|w| match self.last_idx {
        //         None => w.mid_top_of(self.canvas_idx),
        //         Some(ix) => w.down_from(ix, 0.0),
        //     })
        //     .set(self.idx, &mut self.ui);
        item.set(toggle);
    })
    .set(LIST, ui);

```

An option could be provided so that the closure given to the `item` method would only be called for items that are currently visible within the scrollable view for efficiency.

Keep in mind that the above example still assumes a single column, with a single Widget per item. However, note that this could be used as the foundation for multi-column "table"-style widgets as well. I.e. rather than using a `Toggle` as above, a custom `Row` widget could be used which has the necessary number of columns with the desired kinds of widgets (i.e. `TextBox`, `NumberDialer`, etc). I can also imagine a `Table` widget that would wrap the `List` widget and perhaps look something like this:

```rust
Table::new(num_rows, row_h, col_widths)
    .elem(|elem| match elem.col {
        0 => elem.set(Text::new(row_headers[elem.row])),
        1 => elem.set(Button::new().react(|| println!("Foo!"))),
        2 => ...
    })
    .set(TABLE, ui);
```

It should also be possible to simplify the implementations of the `FileNavigator`, `DropDownList` and your `ListView` using the `List` widget above.

I had just about completed my implementation of the `List` idea above earlier this morning, however I realised that the idea is blocked by the generic `Backend` parameter on the `Widget::update` method. The idea above would require that the `Item` struct would hold a `&'a mut UiCell<'a, B>` so that it may instantiate the widgets itself after positioning them (see the comment in the `List` example above). This is currently impossible as the bounds for the closure must take an `Item` arg and would in turn need a `Backend` type param specified in the `impl Widget` signature, which means we'd be unable to specify that the `Backend` type used in the `Item` is the same given via the `Widget::update` method. #734 will remove the `Backend` parameter altogether, so I decided it would be best to focus on landing that first before continuing.

Anyway, hopefully I haven't rambled too much here. Would be curious to get your thoughts on this!

Hyperchaotic:
This looks like a great feature. ListView scratches an immediate itch for myself (also #460 ) and helped me getting to know rust and conrod but as I was duplicating your code for the simple list I was wondering if it could in turn be used back in DirectoryView, thus simplifying that component. I was also wondering how to make a multi column list and share basic code. 

The thing is as you mention a GUI will have several types of lists; a simple list, context menu, a list with multiple (resizeable?) columns with widgets such as checkboxes, icons, etc. and components depending upon them like the file navigator. It's important they all share behavior with regards to selection mechanics which can get a bit hairy. 

BTW; Other things to consider would be to make sure all lists share default theme colors so the UI is consistent and offer similar interfaces to the user. 

I'm wondering, in ListView - as in your example - should we let the user pass in more than just for example strings? If passing in a mutable list of [bool,String] (or [bool,T: Display] ) we could update the selection status directly on user data rather than maintaining an internal copy and maybe simplify the code.

Hyperchaotic:
@tafia Thank you for the feedback! I was wondering what to call a more advanced list view so ListBox is good for this simple one.

mitchmindtree:
@Hyperchaotic now that the `List` widget is added in #758, would you mind refactoring your `ListView` widget to take advantage of it? I've added an example (which I think is quite similar to your use-case, though the multi-select logic will be a bit different) which demonstrates how you can use it.

Also, it looks like you've picked up a lot of unrelated commits when updating your branch to the latest master. Rather than merging with master, would you mind rebasing onto the tip of it instead? You should be able to do this via somthing like:

```
$ git fetch upstream
$ git rebase upstream/master
```

Sorry this is taking a while to land! Thanks for being patient 😸

mitchmindtree:
Perhaps `ListSelection` / `ListSelecter` / `ListSelect`  is a more suitable name, as its main purpose is to allow the user to select different items within the list?

Also, rather than having a `multiple_selections(bool)` method, perhaps we could provide two constructors:
- `ListSelect::single(&list)` and
- `ListSelect::multiple(&list)`

rather than `new`. This way there might be less chance that the user forgets to call `multiple_selection(bool)` and ends up with unexpected behaviour?

Hyperchaotic:
Hiya @mitchmindtree, great news, thank you! I'll look into it this coming week and incorporate your suggestions :)  

I'm new to Git so had no idea how to get my branch updated. I'll rebase.

Hyperchaotic:
Hiya @mitchmindtree I had a look at the list yesterday and it's a breeze using it for single selection list  purposes. However for *multiple selections* where the user holds ctrl or shift it becomes a bit difficult. Pls see below first idea. When a button in the list is pressed I wanted to iterate all widget events for that button (looking for ctrl/shift/etc.) to update selection. However, if I borrow UI in there Rust won't let me use it again in the last line. So I need to find another way of iterating events or getting keyboard status without borrowing ui in the closure.
  
I notice that Item in List have private access to ui, I wonder if List can help. 

Furthermore, if I put the multiple selection code here (in ListSelect), it won't be reuseable for other list implementations.

In fn update(mut self, args: widget::UpdateArgs<Self>)

```
    List::new(self.entries.len() as u32, rect_h)
    .scrollbar_on_top()
    .middle_of(idx)
    .wh_of(idx)
    .item(|item| {
        let i = item.i;
        let label = format!("{}", state.entries[i].item);

        if state.entries[i].is_selected {
            txt_col = sel_text_color; rect_col = sel_rect_color;
        } else {
            txt_col = unsel_text_color; rect_col = unsel_rect_color;
        }
        let ndix = item.widget_idx;
        let button = Button::new()
            .label(&label)
            .label_color(txt_col)
            .color(rect_col)
            .label_font_size(font_size)
            .frame(0.0)
            .react(|| {
                state.update(|state| {state.entries[i].is_selected=!state.entries[i].is_selected;});
                // Borrows UI - but then can't use in .set(list_idx, &mut ui) further down 
                for widget_event in ui.widget_input(ndix).events() {
                    self.update_selection(widget_event, i, previous_selected);
                }
                println!("-{}-", i);
            });
        item.set(button);
    })
    .set(list_idx, &mut ui);

```

mitchmindtree:
Hey @Hyperchaotic, there should be an `Item::ui` method that allows for  borrowing the `Ui` - this should give you access to the events that you need.

Also, rather than trying to check for events inside the `Button`'s `react` closure itself, it might be easier to instead increment a counter and then do the widget event checking after the `Button` is instantiated. This might avoid more ownership issues where the the `Ui` would be borrowed by the closure while the `Item` would require mutable access to it to instantiate the `Button`.

Btw, have you considered using the `Toggle` widget instead of the `Button`? it seems as though it might be more suited to this?

mitchmindtree:
Also, you might know this already, but you can add syntax highlighting to github code-blocks in comments by adding the name of the language next to the back-ticks, i.e.

    ```rust
    code
    ```

Hyperchaotic:
Reimplemented the list selection widget using List and integrated latest changes from master. However when rebasing, git was throwing fatal errors all over the place, it insisted on a merge after the rebase due to diverged branches. I'm new to git (and rust and conrod), so let me know if my branch history is still screwy. If all else fails I can try rebase again now that I'm aligned with master/head, or create a new fork and just move my four changed files over. Don't want to risk polluting anything.

Thank you for your patience!

The ListSelect have the following API:

``` rust 
    /// Construct a new ListSelect, allowing one selected item at a time.
    /// Second parameter is a list reflecting which entries within the list are currently selected.
    pub fn single(entries: &'a Vec<T>, selected: &'a mut Vec<bool>) -> Self

    /// Construct a new ListSelect, allowing multiple selected items.
    /// Second parameter is a list reflecting which entries within the list are currently selected.
    pub fn multiple(entries: &'a Vec<T>, selected: &'a mut Vec<bool>) -> Self
```

The user can thus "pre-select" entries and read updates/changes from the list at runtime.  See examples/list_select.rs
Because there is no internal representation of this list to keep in sync, performance is quite good.
I used Button. Toggle will work for it too but won't make much difference to the slightly hairy selection code I think. 
Selection with SHIFT will now expand an existing block up or down, rather than erasing it and starting over.

Two issues;

- Text is centered, I'd like it to be left justified. Didn't see an option for Button to do that but didn't want to change too much code in this change.
- In examples/list_select.rs - see the commented out "match event" block. I can't get Rust to recognize the Event enum.

mitchmindtree:
> git was throwing fatal errors all over the place

What exactly were the errors? It might have required you to resolve some conflicts which is fairly normal, and following rebasing it might have required you to force push the rebased branch. I'm unsure what other errors you might have run into off the top of my head.

> I can try rebase again now that I'm aligned with master/head, or create a new fork and just move my four changed files over.

Yeah if this is easiest for you that would be great :+1:

Re the new `single`/`multiple` API: I think it would be best to take `&'a [T]` and `&'a mut [bool]` here rather than `Vec`s. `&Vec<T>` is almost never seen in rust as the only benefit of `Vec` over `[T]` is that `Vec` is growable, and seeing as an immutably borrowed `Vec` cannot mutate its size mutated, we might as well pass `&[T]` instead as it is strictly more flexible (also allows fixed size arrays and `Box<[T]>` as these also deref to `&[T]`).

> Text is centered, I'd like it to be left justified. Didn't see an option for Button to do that

Yeah it could be a nice idea for us to add an option for text alignment to the `Button` widget itself. I'm fine with leaving that for a later PR though.

> In examples/list_select.rs - see the commented out "match event" block. I can't get Rust to recognize the Event enum.

What specifically was the compiler error you were getting? It looks like `Event` isn't imported, perhaps that's the issue?

Otherwise, looking good! If you could fix up these things I'd be happy to merge 👍

PR #762, opened 2016-07-31
Add options for List item instantiation

mitchmindtree:
By default, `List` only instantiates visible `Item`s for efficiency.

This PR adds a method that allows for instantiating all `Item`s regardless of their visibility.


mitchmindtree:
Published as **0.38.1**.

PR #763, opened 2016-07-31
issue 689: Added UiBuilder struct

lllShamanlll:
UiBuilder added, allows to create Ui without specifying theme or widgets count.

### DIFF
--- a/src/tests/ui.rs
+++ b/src/tests/ui.rs
@@ -1,5 +1,4 @@
 use {
-    Theme,
     Canvas,
     Color,
     Labelable,
@@ -68,9 +67,7 @@ fn to_window_coordinates(xy: Point, ui: &Ui) -> Point {
 }
 
 fn windowless_ui() -> Ui {
-    let theme = Theme::default();
     UiBuilder::new()
-        .theme(theme)
         .build()
         .unwrap()
 }
--- a/src/tests/ui.rs
+++ b/src/tests/ui.rs
@@ -8,6 +8,7 @@ use {
     Sizeable,
     Widget,
     Ui,
+    UiBuilder
 };
 use event::{self, Input, Motion};
 use input::{self, Button, Key, MouseButton};
@@ -68,7 +69,10 @@ fn to_window_coordinates(xy: Point, ui: &Ui) -> Point {
 
 fn windowless_ui() -> Ui {
     let theme = Theme::default();
-    Ui::new(theme)
+    UiBuilder::new()
+        .theme(theme)
+        .build()
+        .unwrap()
 }
 
 
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -16,7 +16,7 @@ use widget::{self, Widget};
 
 pub struct UiBuilder {
     /// The theme used to set default styling for widgets.
-    theme: Theme,
+    maybe_theme: Option<Theme>,
     /// Number of widgets given as number
     maybe_widgets_capacity: Option<usize>
 }
@@ -104,19 +104,9 @@ impl UiBuilder {
     /// - maybe_widgets_capacity: None
     pub fn new() -> Self {
         UiBuilder {
-            theme: Theme::default(),
+            maybe_theme: None,
             maybe_widgets_capacity: None
         }
-    }
-
-    /// Gets the theme of built **Ui**.
-    pub fn get_theme(&self) -> Theme {
-        self.theme.clone()
-    }
-
-    /// Sets the theme of built **Ui**.
-    pub fn set_theme(&mut self, value: Theme) {
-        self.theme = value;
     }
 
     /// Sets the theme of built **Ui**.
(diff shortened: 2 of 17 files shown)


mitchmindtree:
Hey @lllShamanlll, thanks for the PR!

I just had a look and have a few notes:

- I think it would be best if the `UiBuilder` only has builder methods - there shouldn't be any need for `set`ters or `get`ters.
- The `UiBuilder`'s `theme` field should be an `Option<Theme>` which defaults to `None`. This will save constructing the `Theme` twice on the occasion that the user specifies some custom `Theme`.
- The `UiBuilder::build` method should take `self` rather than `&self` so that we don't require `clone`ing its contents. This will also allow you to change back to using `Box<Any>` in the `Theme`'s `widget_style` `HashMap` and remove the `Clone` implementation.
- If we're going to use the `UiBuilder` to construct the `Ui`, we should make the `Ui::new` constructor private so that users don't get confused between multiple ways of constructing the `Ui`.

lllShamanlll:
@mitchmindtree  thx for review!
About setters/getters  i've just tried to be consistent with builders in piston.
I'll fix this, but what about getting builder data from Ui::new?
D'u suggest pass all building data as arguments to Ui::new? Doesn't it make Ui::new unscalable?
Or completely remove Ui::new in favor of creating Ui completely on UiBuilder::build side?

mitchmindtree:
No problem!

> About setters/getters i've just tried to be consistent with builders in piston.

I'd prefer we do not include `get`ters and `set`ters as it adds ambiguity to the way that the `UiBuilder` should be used. It should solely be used to construct the `Ui` using the builder pattern.

> I'll fix this, but what about getting builder data from Ui::new?

When specifying the `theme` field of the `Ui` constructor, do this: 
```rust
theme: theme.unwrap_or_else(Theme::default),
```
This will use the `UiBuilder`'s theme if one was specified or constructs the default if it was `None`.

> D'u suggest pass all building data as arguments to Ui::new? Doesn't it make Ui::new unscalable?
> Or completely remove Ui::new in favor of creating Ui completely on UiBuilder::build side?

I'm not exactly sure what you mean by this. I don't mind whether we keep `Ui::new` as a private constructor that the user cannot see (currently [it is still `pub` in your PR](https://github.com/PistonDevelopers/conrod/pull/763/files#diff-46f6f517ae95d07cddf159932baae1a7R176)) or if we inline it into the `UiBuilder::build` method itself. The important thing is that the only way a user should be able to build a `Ui` is using the `UiBuilder::build` method.

src/tests/ui.rs
```diff
@@ -68,7 +69,10 @@ fn to_window_coordinates(xy: Point, ui: &Ui) -> Point {
 
 fn windowless_ui() -> Ui {
     let theme = Theme::default();
-    Ui::new(theme)
+    UiBuilder::new()
+        .theme(theme)
```
mitchmindtree: Remember this is no longer necessary as `Theme::default()` is now the default.

src/ui.rs
```diff
@@ -88,36 +97,62 @@ pub struct UiCell<'a> {
...
-                    depth_order: graph::DepthOrder,
-                    updated_widgets: std::collections::HashSet<NodeIndex>) -> Self
-    {
         let window = widget_graph.add_placeholder();
         let prev_updated_widgets = updated_widgets.clone();
-        Ui {
+        Ok(Ui {
             widget_graph: widget_graph,
-            theme: theme,
+            theme: builder.maybe_theme.unwrap_or(Theme::default()),
```
mitchmindtree: This line always constructs the `Theme::default()`, even if `maybe_theme` is `Some`.

Instead, a more efficient alternative here would be: 
```rust
theme: builder.maybe_theme.unwrap_or_else(Theme::default),
```

Here, the function `Theme::default` is instead only called if `maybe_theme` is `None`.

src/ui.rs
```diff
@@ -88,36 +97,62 @@ pub struct UiCell<'a> {
 /// buffer. Otherwise if we don't draw into each buffer, we will probably be subject to flickering.
 pub const SAFE_REDRAW_COUNT: u8 = 3;
 
+impl UiBuilder {
+    /// Create **Ui** builder with defaults.
+    ///
+    /// - theme: Theme::default()
```
mitchmindtree: Should this be changed to `maybe_theme`?

src/ui.rs
```diff
@@ -88,36 +97,62 @@ pub struct UiCell<'a> {
...
+            maybe_theme: None,
+            maybe_widgets_capacity: None
+        }
+    }
 
-impl Ui {
+    /// Sets the theme of built **Ui**.
+    ///
+    /// This method moves the current builder,
+    /// unlike [`set_theme()`](#method.set_theme),
```
mitchmindtree: There is no longer any `set_theme` method, perhaps these doc comments should be updated?

PR #766, opened 2016-08-01
Export `Item` type from list module. Add method for borrowing the `UiCell` from `ListItem`.

mitchmindtree:


mitchmindtree:
Published as **0.38.2**.

Issue #729, opened 2016-06-07
`TextEdit` todo list


mitchmindtree:
#703 adds a multi-line `TextEdit` widget, but it still requires a lot of work:

- [ ] Insert a newline on `Return`/`Enter`.
- [ ] Add functionality for `Up` and `Down` arrow keys.
- [ ] Add clipboard functionality.
- [ ] Fix bug where cursor sometimes jumps to the start of the line when entering text at the end of a line.
- [ ] Fix issue where `Text` widget is slightly offset from drawn selection rectangles when `TextEdit` is aligned to the middle of bottom of the *y* axis. Doesn't seem to be a problem when aligned to the top.

christolliday:
So with https://github.com/PistonDevelopers/conrod/issues/796 I think the only issue left is the clilpboard.

I think the best option for cross-platform clipboard support is to depend on https://github.com/aweinstock314/rust-clipboard. The only problem is parts of it are GPLv2, but apparently there are some plans to rewrite those parts, so maybe the easiest thing to do is to wait for that. I think otherwise we would have to relicense as GPL?

akhilman:
https://github.com/chrisduerr/rust-clipboard have Apache and MIT license.

PR #767, opened 2016-08-02
Fix `TextEdit` cursor jump bugs

tmerr:
Fix bugs in `TextEdit` where the cursor can jump to unexpected locations
(usually the upper left of the screen) while editing. This could happen
when pressing the left arrow key or backspace at the beginning of a line,
or when pressing space at the end of a line.

The causes were minor mistakes like off-by-one errors. Related: #729

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -503,7 +503,7 @@ pub mod cursor {
                 let new_line = line - 1;
                 line_infos.nth(new_line)
                     .map(|info| {
-                        let new_char = info.end_char();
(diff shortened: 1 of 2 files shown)


mitchmindtree:
@tmerr thanks for the PR! I think I know the bug you're talking about, I'll test this locally now :+1:

mitchmindtree:
Great job spotting these bugs! Hugely appreciated 😸 

Everything works fine for me, merging :+1:

PR #770, opened 2016-08-02
Bugs/689 `Theme` made optional for `Ui`

lllShamanlll:
Added struct `UiBuilder`, able to build new `Ui` without theme/widget capacity specified.

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -128,7 +128,7 @@ impl UiBuilder {
     }
 
     /// Build **Ui** from the given builder
-    pub fn build(self) -> Result<Ui, String> {
+    pub fn build(self) -> Ui {
         Ui::new(self)
     }
 }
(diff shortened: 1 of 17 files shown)


src/ui.rs
```diff
@@ -88,36 +97,60 @@ pub struct UiCell<'a> {
...
+    /// Build **Ui** from the given builder
+    pub fn build(self) -> Result<Ui, String> {
+        Ui::new(self)
     }
+}
+
+impl Ui {
+
+    /// A new, empty **Ui**.
+    fn new(builder: UiBuilder) -> Result<Self, String> {
```
mitchmindtree: I don't think we need to return a `Result` here or in `UiBuilder::build`

mitchmindtree:
Looking good :+1:

Thanks a lot for addressing all my comments, sorry if it got a bit tedious!

PR #771, opened 2016-08-02
Bugs/768 Supported non-default window backend for PistonWindow in piston_window.rs

lllShamanlll:


mitchmindtree:
Thanks!

In the future, I recommend writing more descriptive commit messages. It can get very hard to navigate git logs with messages like this one.

PR #773, opened 2016-08-04
Add a new `render::OwnedPrimitives` type as an owned alternative to the `render::Primitives` type. Useful for threading the UI.

mitchmindtree:
I started on this after overflowing the stack on main in a personal project with a particularly large GUI. In conrod the user builds new widgets by instantiating other widgets within the `Widget::update` method. This normally isn't an issue but can lead to very large stack sizes when the widget hierarchy gets very deep. Rust currently does not provide a portable way of requesting a greater stack size for the main thread, however it does for spawning child threads. This lead me to want to instantiate the `Ui` on a dedicated thread (with a larger stack size), which in turn lead me to requiring an "owned" version of the `Primitives` type so that I could still send rendering data to main.

This also removes the `image::Map` field from the Primitives type in favour of passing it directly to the backend as necessary. This simplifies cases where the `Ui` is running on a thread other than main. As a result however, there is no longer any way for the `image::Map` to notify the `Ui` of when a redraw is necessary due to some mutation of state. For now the user is required to manually check whether redraw is required via the `image::Map`'s `trigger_redraw` field (which will also need to be manually reset after drawing) until we can come up with a nicer automated solution.

I also added new functions to the `piston` and `piston_draw` backends for drawing individual `Primitive`s.

### DIFF
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -180,10 +180,11 @@ fn main() {
         //
         // If instead you need to re-draw your conrod GUI every frame, use `Ui::draw`.
         window.draw_2d(&event, |c, g| {
-            if let Some(primitives) = ui.draw_if_changed(&image_map) {
+            if let Some(primitives) = ui.draw_if_changed() {
                 fn texture_from_image<T>(img: &T) -> &T { img };
                 conrod::backend::piston_window::draw(c, g, primitives,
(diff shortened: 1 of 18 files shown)


mitchmindtree:
Just finished testing the new `OwnedPrimitives` type on a couple personal projects and it seems to be working nicely.

PR #778, opened 2016-08-05
Created ListSelect widget.

Hyperchaotic:
I did a rebase --hard (which destroyed the old pull req) and then added the ListSelect. Branch should be OK now.

Have replaced Vec with [T] and [bool], it looks cleaner, thanks!

It works fine. Only thing is the Event enum. I can't seem to export/import it for the example. It's not in conrod::Event and ListSelect does not have it's own module. I can't import it as 'use conrod::widget::list_select::Event' as Rust says widget is private.

### DIFF
--- a/src/widget/list_select.rs
+++ b/src/widget/list_select.rs
@@ -1,0 +1,465 @@
+
+use {
+    color,
+    Color,
+    Colorable,
+    event,
+    FontSize,
+    IndexSlot,
+    Labelable,
+    NodeIndex,
+    Positionable,
+    Scalar,
+    List,
+    Button,
+    Borderable,
+};
+use widget::{self, Widget};
+use std::fmt::Display;
+
+use std;
+
+/// Displays a given Vec<T> where T: Display as a selectable List. Its reaction is triggered upon
+/// selection of a list item.
+pub struct ListSelect<'a, T, F> where F: FnMut(Event), T: Display+'a {
+    entries: &'a [T],
+    selected: &'a mut [bool],
+    common: widget::CommonBuilder,
+    maybe_react: Option<F>,
+    multiple_selections: bool,
+    style: Style,
+}
+
+widget_style!{
+    /// Styling for the ListSelect, necessary for constructing its renderable Element.
+    style Style {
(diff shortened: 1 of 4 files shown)


mitchmindtree:
@Hyperchaotic nice one! The diff is looking much cleaner, thanks a lot for that 😸 

Re `Event`, I normally re-export widget-specific events from the top level so that they can be accessed, i.e in lib.rs, add something like this:

```rust
pub use widget::list_select::Event as ListSelectEvent;
```

Hopefully you'll be able to see this pattern when you have a look at the lib.rs. This will let you import `ListSelectEvent` in your example, and match on it like this:

```rust
match event {
    ListSelectEvent::SelectEntry => // do stuff,
    ....
}
```

As a side note: Eventually I'd like to try and remove a lot of the top-level re-exports in favour of making modules like `widget` public, but I still need to work out how to keep certain api's private before making that change.

*Related to #749*

src/widget/list_select.rs
```diff
@@ -0,0 +1,465 @@
...
+
+pub use self::Event::{SelectEntry, SelectEntries, DoubleClick, KeyPress};
+
+/// The kind of events that the `ListSelect` may `react` to.
+/// Provides tuple(s) of index in list and string representation of selection
+#[derive(Clone, Debug)]
+pub enum Event {
+    SelectEntry((usize, String)),
+    SelectEntries(Vec<(usize, String)>),
+    DoubleClick((usize, String)),
```
mitchmindtree: It might be a little more idiomatic to change this variant to
```rust
DoubleClick(usize, String),
```
rather than using a tuple (same goes for the `SelectEntry` variant)?

Hyperchaotic:
Oh they're exported in lib.rs (facepalm).... I've incorporated your comments. It compiles as debug and release and I've run cargo test. I hope this'll satisfy issue #460    :smile:

mitchmindtree:
Woohoo!

Looks great @Hyperchaotic, thanks for the contribution and for putting up with my comments! I might make a couple more small changes later on (will cc you of course) but I'm happy to merge this as it is :+1:

PR #785, opened 2016-08-10
Images on buttons #171

pierrechevalier83:
Minimum working fix for #171

* A texture can now be added to a button.
* An example is added to test/demonstrate how to use it.

### DIFF
--- a/examples/button.rs
+++ b/examples/button.rs
@@ -1,0 +1,159 @@
+//!
+//! A demonstration of all non-primitive widgets available in Conrod.
+//!
+//!
+//! Don't be put off by the number of method calls, they are only for demonstration and almost all
+//! of them are optional. Conrod supports `Theme`s, so if you don't give it an argument, it will
+//! check the current `Theme` within the `Ui` and retrieve defaults from there.
+//!
+
+#[macro_use] extern crate conrod;
+extern crate find_folder;
+extern crate piston_window;
+extern crate rand; // for making a random color.
+
+use piston_window::{EventLoop, PistonWindow, UpdateEvent, WindowSettings};
+
+
+/// This struct holds all of the variables used to demonstrate application data being passed
+/// through the widgets. If some of these seem strange, that's because they are! Most of these
+/// simply represent the aesthetic state of different parts of the GUI to offer visual feedback
+/// during interaction with the widgets.
+struct Context {
+    /// Background color (for demonstration of button and sliders).
+    bg_color: conrod::Color,
+}
+
+impl Context {
+    /// Constructor for the Demonstration Application model.
+    fn new() -> Context {
+        Context {
+            bg_color: conrod::color::rgb(0.2, 0.35, 0.45),
+        }
+    }
+
+}
+
+
+fn main() {
+    const WIDTH: u32 = 1100;
+    const HEIGHT: u32 = 560;
+
+    // Change this to OpenGL::V2_1 if not working.
+    let opengl = piston_window::OpenGL::V3_2;
+
+    // Construct the window.
+    let mut window: PistonWindow =
+        WindowSettings::new("A button with an image", [WIDTH, HEIGHT])
+            .opengl(opengl).exit_on_esc(true).vsync(true).build().unwrap();
+
+    // construct our `Ui`.
+    let mut ui = conrod::UiBuilder::new().build();
+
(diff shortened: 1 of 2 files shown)


mitchmindtree:
@pierrechevalier83 nice one! Thanks for going ahead with this, looking good! Just a couple notes:

- I think we might be better off implementing this as an inherent method, I don't think there's any need to add a new trait just yet. The main reason the other traits exist (i.e. `Colorable`, `Borderable`, etc) is because they provide multiple useful methods while only requiring that the user implements one. I'm still not sure how I feel about them either, as they can be a pain to import and we never actually require being generic over any of them. I can tell you were probably just trying to follow existing convention here though, and I appreciate that!
- I think `image` might be a better name for the method, as we're not technically limited to using textures here.
- Rather than taking a `widget::Index` directly, perhaps the `image` method could take some `I` where `I: Into<widget::Index>`? This way the method can be generic over either `widget::Id`, `widget::Index` or `graph::NodeIndex` which will remove the need to write things like `widget::Index::from(RUST_LOGO)` as in the new example.

examples/button.rs
```diff
@@ -0,0 +1,158 @@
...
+/// allocations by updating the pre-existing cached state. A new graphical `Element` is only
+/// retrieved from a `Widget` in the case that it's `State` has changed in some way.
+fn set_widgets(ui: &mut conrod::UiCell, app: &mut Context) {
+    use conrod::{color, widget, Colorable, Positionable, Sizeable, Texturable, Widget};
+
+    // We can use this `Canvas` as a parent Widget upon which we can place other widgets.
+    widget::Canvas::new()
+        .pad(30.0)
+        .color(app.bg_color)
+        .scroll_kids()
```
mitchmindtree: We probably don't need to call this as there aren't enough widgets to cause the `Canvas` to become scrollable.

examples/button.rs
```diff
@@ -0,0 +1,158 @@
...
+        .set(CANVAS, ui);
+    // Button widget example button.
+    widget::Button::new()
+        .w_h(200.0, 200.0)
+        .middle_of(CANVAS)
+        .rgb(0.4, 0.75, 0.6)
```
mitchmindtree: If the `Button` is using an `Image`, does this coloring method do anything? Hmmm, we should work out some way so that if the user has specified an image, they either:
- cannot call `Colorable` methods
- if they do call `Colorable` methods, this sets the `color` method on the `Image` used inside the `Button`.

I have some ideas, but would probably be easier for me to do a follow up PR and CC you in it rather than explain them here.

pierrechevalier83: > If the Button is using an Image, does this coloring method do anything?
If I undestand properly, yes actually:
If the image is a PNG with transparency, the color will be visible in the background.
Please, tell me if I'm missing something here.

examples/button.rs
pierrechevalier83: This makes sense: I'll fix this.

pierrechevalier83:
Hi @mitchmindtree , thanks for the review.

About your first point: I went with a trait both because I was following the convention and because I was thinking that it may make sense to implement the trait on kind of any widget that can have a Color.
I'm happy with reverting back to a simple method, though: keep it simple until complexity is actually needed.

About your second point: yes, texture was a misnommer. Mostly I couldn't bring myself to write the word: "Imageable" :) Will use image.

About your third point: great suggestion! When writing the example, I was thinking to myself this was a bit clumsy; but I didn't think of this solutil. Will do.

examples/button.rs
mitchmindtree: > If the image is a PNG with transparency, the color will be visible in the background.

Ahh I see what you mean, yes this would be the case.

I believe `Image`s can be colored themselves, so there might be some ambiguity here in the behaviour expected by the user. For example, if a user was using an image for their button, they might expect to see the image *instead* of the flat color rather than on top of it, and to have the image colored via its luminance (I think this is what piston graphics does, for example). I wonder if we could distinguish between these behaviours somehow. Thoughts?

pierrechevalier83: The way I see it:
The image on a button doesn't have to be as versatile as the one in an image widget.
The user doesn't have access to the underlying Image widget, so they can't set its color or precise position within the widget and that's not a big problem. If they want full control, they probably want to write a custom widget for their use case.

Not sure...

Issue #746, opened 2016-07-08
Add a method to `FileNavigator` for optionally displaying hidden files and folders.


mitchmindtree:

Hyperchaotic:
I did some preliminary research tonight, pondering on the differences between platforms for this. On Unix like platforms (Linux, OS X) it's straightforward of course as a dot file is hidden. However on Windows we need to look for the dwFileAttributes field for the file. This is available as an [unstable API here](https://aochagavia.github.io/rust/std/os/windows/fs/trait.MetadataExt.html).

Example code:

``` rust
            use std::os::windows::fs::MetadataExt;

            const WIN_ATTR_HIDDEN: u32 = 0x2;

            let metadata = fs::metadata(&path).unwrap();
            let win_attr: u32 = metadata.file_attributes();

            print!("File {:?}, attr {:b} - ", path, win_attr);
            if (win_attr & WIN_ATTR_HIDDEN)!=0 {
                println!("is hidden! ");
            } else {
                println!("is NOT hidden! ");
            }

```
I ran the code (compiled with nightly) over a few files, one marked as hidden, and got this:
```
File "breakdance.avi", attr 10000000100000 - is NOT hidden!
File "breakdance-hidden.srt", attr 10000000100010 - is hidden!
```
The code would need to be guarded by compiler directives, [like these](https://doc.rust-lang.org/book/conditional-compilation.html) I guess.

mitchmindtree:
This was implemented in #789, closing.

PR #789, opened 2016-08-15
Implemented issue 746: Add a method to `FileNavigator` for optionally...

Hyperchaotic:
… displaying hidden files and folders.

FileNavigator now defaults to not showing hidden files and directories. Have to be enabled with the show_hidden_files builder method.

Tested successfully on Linux Mint and Windows. Couldn't get my trusty 2006 MacBook Pro to install Rust due to Snow Leopard :(   

The implementation uses the unstable API mentioned in #746, I assumed we're using the nightly compiler.

### DIFF
--- a/src/widget/file_navigator.rs
+++ b/src/widget/file_navigator.rs
@@ -32,6 +32,8 @@ pub struct FileNavigator<'a> {
     pub starting_directory: &'a std::path::Path,
     /// Only display files of the given type.
     pub types: Types<'a>,
+    /// Whether to show hidden files and directories
+    show_hidden: bool,
 }
 
 /// A type for specifying the types of files to be shown by a `FileNavigator`.
@@ -117,6 +119,7 @@ impl<'a> FileNavigator<'a> {
             style: Style::new(),
             starting_directory: starting_directory,
             types: types,
+            show_hidden: false,
         }
     }
 
@@ -143,6 +146,13 @@ impl<'a> FileNavigator<'a> {
     /// The color of the `Text` used to display the file names.
     pub fn text_color(mut self, color: Color) -> Self {
         self.style.text_color = Some(Some(color));
+        self
+    }
+
+    /// Whether to show hidden files and directories. On Windows a hidden file is identified
+    /// by a file attribute flag, on other OSs the file name starts with a '.'.
+    pub fn show_hidden_files(mut self, show_hidden: bool) -> Self {
+        self.show_hidden = show_hidden;
         self
     }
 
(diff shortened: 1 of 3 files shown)


mitchmindtree:
> The implementation uses the unstable API mentioned in #746, I assumed we're using the nightly compiler.

Ahh that's a pity that this would require nightly on Windows. Is there any chance we could perhaps lift the unstable code into a module temporarily so that we can remain on stable? Or perhaps there is a temporary stable solution on crates.io that we can use in the meantime?

Conrod has been building on stable since rustc 1.0.0 was released and I think it's important that we continue to do so. I haven't used nightly since pre-1.0.0 as the constant breakage before rust went stable used to cost a *lot* of time. As someone who maintains a lot of libraries while working on their own large rust project, it's just not worth the chance as even minor breakages can cost days to adapt to across many crates.

Hyperchaotic:
Oh yeah, I absolutely agree there's no need to change compiler for a tiny feature. I would agree to wait, to keep things simple and just always show all files on windows for now. 

I might first have a look for compiler directives, perhaps there's a flag for compiler type, API availability or even we could define a custom feature flag to enable at compilation time:
`
cargo build --features "windows_file_attr"
`

Hyperchaotic:
I've replaced it with a stub function on windows. The original can be enabled in the build by adding/enabling the feature "windows_metadataext". We can replace the stub and make it default when the API becomes stable.

Before implementation I did search for crates for alternatives to read_dir that accounts for hidden files on all platforms, but didn't find any that were truly cross platform.

src/widget/file_navigator.rs
```diff
@@ -483,6 +495,45 @@ pub mod directory_view {
...
+            return (win_attr & FILE_ATTRIBUTE_HIDDEN) != 0;
+        }
+        false
+    }
+
+    #[cfg(not(target_os = "windows"))]
+    /// Check if a file is hidden on any other OS than windows, using the dot file namings.
+    fn is_file_hidden(path: &std::path::PathBuf) -> bool {
+        let name = path.file_name();
+        println!("Linux metadataext");
```
mitchmindtree: Looks like a sneaky debug `println` made it through

src/widget/file_navigator.rs
```diff
@@ -483,6 +495,45 @@ pub mod directory_view {
...
+        if let Some(name) = name {
+            return name.to_string_lossy().to_owned().starts_with(".");
```
mitchmindtree: I think `.to_owned()` might be unnecessary here?

mitchmindtree:
Nice one @Hyperchaotic, this is great :) If you could address those two little comments i'd be happy to merge!

Hyperchaotic:
Thank you for the review @mitchmindtree ! I hadn't realized Cow implements Deref, it's nice. Also removed two warnings in list_select.rs.

PR #793, opened 2016-08-15
TextEdit up/down, return and fixed missing cursor on empty lines

christolliday:
This adds the features of moving up and down in a TextEdit via the arrow keys, and inserting a new line when you press return. This ticks off two items on https://github.com/PistonDevelopers/conrod/issues/729

I did a little refactoring to simplify the implementation on those two features. I removed the closure that maps the text onto the lines and moved it into the iterator itself, in order to be able to return an XysPerLine iterator. The return type on xys_per_line_from_text is pretty ugly but I'm hoping that could be cleaned up once abstract return types are supported, couldn't think of a better way than returning that complex iterator.

Also fixed two small bugs, first, when the text ends in a new line, ie.  'abc\n', it should now display two lines rather than one:
    abc
    |
(the | is the cursor)

Second, fixed the cursor not displaying correctly when the text is empty. When the TextEdit is center aligned, and the text is empty, it would show a left aligned cursor that would then jump to the center after you start typing. I fixed these two by having the Infos iterator return an empty line Info for empty lines, instead of None.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -803,6 +803,8 @@ pub mod line {
         start_byte: usize,
         /// The character index that indicates the start of the next line to be yielded.
         start_char: usize,
+        /// The break type of the previously yielded line
+        last_break: Option<Break>,
     }
 
     /// An iterator yielding a `Rect` for each line in 
@@ -864,6 +866,7 @@ pub mod line {
                 next_break_fn: self.next_break_fn.clone(),
                 start_byte: self.start_byte,
                 start_char: self.start_char,
+                last_break: None,
             }
         }
     }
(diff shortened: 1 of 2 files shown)


mitchmindtree:
@christolliday amazing! Thanks so much for this, I just had a play with your fork and it seems to work nicely :+1: I'll have a closer review of the diff now

mitchmindtree:
This all looks great to me ❤️ 

> The return type on xys_per_line_from_text is pretty ugly but I'm hoping that could be cleaned up once abstract return types are supported

Yeah, I think for now it might be best to wrap this in a new `XysPerLineFromText` Iterator type, but I'm happy to do this myself. It seems like `impl Iterator` will remove the need for a third of the code in that text.rs module - that feature has been a pipe dream for so long that I'm not going to know what to do with myself once it lands :)

I'm happy to merge this!

PR #794, opened 2016-08-16
Move `Image` option for `Button` from runtime to compile time with a type parameter. Add image coloring options.

mitchmindtree:
Move the image support along with related methods into their own type parameter to try and properly distinguish behaviour at compile time. Follow up to discussion in #171.

Adds coloring options for Image.

Also fixes formatting issues by replacing tabs introduced in #171 with spaces.

cc @pierrechevalier83


pierrechevalier83:
Thanks for that @mitchmindtree. Sorry: maybe I pushed my version a bit fast. I'll play with this code for a while and give you feedback if I have any.

PR #796, opened 2016-08-16
TextEdit move by word, move to end, fix layout bugs

christolliday:
Implemented Ctrl+E to move to end of the text, and Ctrl+Left/Ctrl+Right to move by word through the text.

I think this also fixes the 5th item in https://github.com/PistonDevelopers/conrod/issues/729

Finally, fixed a bug with line breaking by new line, when a word is longer than max_width, it would insert empty lines rather than breaking up the oversized word.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -512,15 +512,18 @@ pub mod cursor {
             let Index { line, char } = self;
             if char > 0 {
                 line_infos.nth(line).and_then(|line_info| {
+                    let line_count = line_info.char_range().count();
                     let mut chars_rev = (&text[line_info.byte_range()]).chars().rev();
+                    if char != line_count {
+                        chars_rev.nth(line_count - char - 1);
+                    }
                     let mut new_char = 0;
                     let mut hit_non_whitespace = false;
-                    chars_rev.nth(line_info.char_range().count() - char);
                     for (i, char_) in chars_rev.enumerate() {
                         // loop until word starts, then continue until the word ends
                         if !char_.is_whitespace() { hit_non_whitespace = true; }
                         if char_.is_whitespace() && hit_non_whitespace {
-                            new_char = char - (i + 1);
+                            new_char = char - i;
                             break
                         }
                     }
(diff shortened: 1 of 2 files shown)


christolliday:
So I just found this issue https://github.com/PistonDevelopers/conrod/issues/71 after the fact, and the discussion here https://github.com/PistonDevelopers/conrod/pull/559
it seems there is some confusion over Ctrl+Left/Ctrl+Right behaviour. For me in Linux and Windows it usually moves by word rather than to the beginning and end of the line.
It seems like ideally there should be some configurable mapping of controls, with different defaults for different platforms.

src/text.rs
```diff
@@ -497,6 +497,75 @@ pub mod cursor {
...
+                })
+            } else {
+                self.previous(line_infos)
+            }
+        }
+
+        /// The cursor index of the end of the first word (block of non-whitespace) after `self`.
+        ///
+        /// If `self` is at the end of the line, call previous, which returns the last
+        /// index position of the previous line, or None if it's the first line
```
mitchmindtree: Is this correct? Looks like it might still need to be changed after copy pasta?

src/text.rs
```diff
@@ -497,6 +497,75 @@ pub mod cursor {
...
+        ///
+        /// If `self` points to whitespace, skip past that whitespace, then return the index of 
+        /// the end of the word after the whitespace
+        ///
+        /// If `self` is in the middle or start of a word, return the index of the end of that word
+        pub fn next_word<I>(self, text: &str, mut line_infos: I) -> Option<Self>
```
mitchmindtree: Perhaps it could be worth making these new method names a little more explicit, i.e. `end_of_next_word` and `start_of_previous_word`?

mitchmindtree: Or maybe `next_word_end` and `next_word_start` might be nicer?

mitchmindtree:
I left a couple comments, otherwise this looks great!

> Ctrl+Left/Ctrl+Right behaviour ... in Linux and Windows it usually moves by word

On OS X `Alt` gives the behaviour you describe. Curiously, both `Alt+Up` and `Ctrl+Left` jump to the start of the line while `Alt+Down` and `Ctrl+Right` jump to the end.

> It seems like ideally there should be some configurable mapping of controls

Yeah agreed, perhaps we can can add these mappings to the `TextEdit`'s `Style` so that the user can set their own defaults within the `Theme` if they wish. I'm imagining adding something like this to the `widget_style!` macro invocation:
```rust
/// When held, this key causes the cursor to move to the end of the next word.
- move_to_word_end_modifier: input::ModifierKey { default_move_to_word_end_modifier() }
/// When held, this key causes the cursor to move to the end of the next line.
- move_to_line_end_modifier: input::ModifierKey { default_move_to_line_end_modifier() }
```

Where `default_move_to_{word/line}_end_modifier` are functions that return the platform default modifier. Thoughts?

src/text.rs
christolliday: ah yep good catch, I'll fix that

src/text.rs
christolliday: Ok, yeah I like next_word_end, previous_word_start, it's more explicit and still consistent with previous, next

christolliday:
Interesting, yeah I think using styles and themes for this makes sense. I suppose the one downside to that is just in naming, that Style then becomes more of a general configuration, than purely for visual style, but that's nothing serious.
I think just configuring the modifier keys could be too limiting though, as opposed to configuring the key combination, as in, the key and set of modifiers. There should probably be some struct for a key binding for that, does anything like that exist in Piston input?

christolliday:
ok fixed the problems you addressed, also just fixed a small bug I missed in next_word_end, previous_word_start.
The issue was for previous_word_start, `word w|ord` would move to `|word word`, rather than `word |word`, and similarly for next_word_end

mitchmindtree:
Looks good! I'm happy to merge this as is, I'll add a point to the `TextEdit` for configuring key combinations.

PR #803, opened 2016-08-20
Make Widget::State and Style Send

flodiebold:
I'd like to use the Ui from another thread sometimes, but it's not Send because of the various Anys. How about requiring Send for Widget::State and Style? It doesn't look like it should cause any problem, but I don't know anything about the conrod internals...

### DIFF
--- a/src/graph/mod.rs
+++ b/src/graph/mod.rs
@@ -74,7 +74,7 @@ pub struct UniqueWidgetState<State, Style> where
 #[derive(Debug)]
 pub struct Container {
     /// Dynamically stored widget state.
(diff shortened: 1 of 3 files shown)


mitchmindtree:
@flodiebold Interesting! I'm currently running conrod on a unique thread, but haven't come across the need to send it across threads after construction yet - would you mind describing your use case for this?  This looks fine to me btw, but I'd just like to give it a little more thought before merging

flodiebold:
I'm building the command buffers for my 3D rendering on another thread, and want to do the UI rendering there, too. It will probably not be relevant for performance, it's mainly for consistency :)

Issue #140, opened 2014-10-15
`Theme` checklist.


mitchmindtree:
Basic `Theme`s are now implemented #139, but there's still a lot of stuff left to do.
- [ ] ~~json serialization~~ easier customisation/configuration (either some sort of deserialization or perhaps a nice macro).
- [ ] user friendly way of creating / editing themes without touching code (maybe something like `.yml`?).
- [x] add the option to specify unique styling for each widget rather than just one general style.
- [x] include default widget dimensions in the theme.
- [x] come up with a more elegant way of retrieving style data in widget modules.
- [ ] an example of loading, saving and creating custom themes.

If you think of something feel free to add a comment and I'll add it to the list.

Spencer-H:
Is yml the preferred theme format? Or would it be reasonable to use CSS, perhaps the crates from Servo for dealing with CSS parsing?

mitchmindtree:
Hey @Spencer-H, sorry again for missing this and thanks for trying to push this discussion along!

I think fully supporting CSS might be overkill for conrod, however I personally don't have enough experience with it to see how we would map it to the `Theme` type.

`.yml` as you mentioned could be a nice possibility. However, if we do opt for de-serialization as the way to write custom `Theme`s, I think my personal preference would be to use the `.toml` format, seeing as most rustaceans should be familiar with it due to it's use with cargo.

That said, I'm currently leaning towards writing a macro to solve custom themes. Some benefits would include:
- We could make its syntax as ergonomic (or at least close to) the other serialization formats mentioned above.
- We wouldn't require any 3rd party crate dependencies
- The custom `Theme` could be loaded at compile time rather than from a separate file during run time.
- The user doesn't have to leave rust or learn any other configuration formats.

Would be great to get your thoughts if you're still interested!

Spencer-H:
It's fine. And as for macros and toml, I assumed that themes would be meant
to be something dropped in for a whole system like Qt and GTK themes, I
didn't realize we meant for them to be application specific and static. In
that case though, a macro makes plenty of sense.

louy2:
+1 for CSS. Maybe also a step in the direction of a XML based layout description language that's all the hype these days in GUI dev, from WAML.C# to HTML.Electron to QML.QtQuick.

drunkwcodes:
There is a Rust implementation of CSS Syntax Level 3.

https://github.com/servo/rust-cssparser

ammon0:
[...]

PR #807, opened 2016-08-26
JSON Serialization for Theme

Spencer-H:
This also required implementing it for some other types.

Some things involving Widget ID's and Font ID's were not implemented because I was unsure how to represent those between applications or runs. 
#140

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ daggy = "0.4.0"
 num = "0.1.30"
 pistoncore-input = "0.13.0"
 rusttype = "0.2.0"
+json = "0.10.2"
 
 # Optional dependencies and features
 # ----------------------------------
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,6 +12,7 @@ extern crate daggy;
 extern crate num;
 extern crate input as piston_input;
 extern crate rusttype;
+#[macro_use] extern crate json;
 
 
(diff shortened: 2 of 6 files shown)


mitchmindtree:
@Spencer-H thanks for the PR!

Sorry for the vagueness in the original `Theme` checklist at #140 - I should have updated it with specifics and other details that I'd like to see solved before going ahead with it.

The idea behind having `Theme` serialization would be to make custom `Theme` configurations easier for users to write. I'm not sure that converting *to* JSON would be very useful, as opposed to de-serializing *from* JSON.

Some other unanswered questions I'd like to answer before going ahead with anything like this:
- Is JSON really the best configuration format for us to use?
- Perhaps we should consider `.toml` as it seems friendlier to "manual" configuration and is perhaps more conventional for configuration in the rust ecosystem (this is what cargo uses for example).
- Perhaps we should leave it up to the user to decide how they wish to serialize `Theme`s, and instead provide a generic `serde` `Deserialize` implementation?
- Maybe we can avoid serialization altogether by providing a friendly macro for custom `Theme`s?

Thanks again for the PR, but I'm going to close this for now until we can answer these questions. 

Agh I just noticed that you left a comment querying about this a while ago! I'm sorry that I never got around to responding, it's difficult to keep track of discussion across all of the projects I'm contributing too 😕 I updated the checklist point at #140 to clarify what I meant. I'll respond to your original comment now.

PR #810, opened 2016-09-01
Simplify widget identifiers throughout conrod - Only use `widget::Id` [Breaking Change]

### LINKED ISSUE #791: Privatise the `widget::Id`, `graph::NodeIndex` and `IndexSlot` types in favour of only exposing `widget::Index`
@tmerr and I were discussing the various types of IDs and indices used throughout conrod and the complexity that is imposed upon users as a result. We ended up realising that all of these roles could probably be encapsulated in the variants of the `widget::Index` and that there shouldn't really be any need for the user to be aware of the `widget::Id`, `graph::NodeIndex` or `IndexSlot` types at all. Hopefully this can simplify things from the user perspective, and even remove some of the generics (`I: Into<Widget>`) used through conrod APIs.

mitchmindtree:
Simplify widget identifiers throughout conrod - Only use `widget::Id`

Recently there has been some confusion around the number of different
identifiers and related types used throughout conrod. Most of these
types revolve around allowing conrod to support both compile-time
generated identifiers (`widget::Id`s) and run-time generated identifiers
(`NodeIndex`s). #791 summarises the pre-cursor discussion to this commit
highlighting that it is becoming too complex to support both of these
kinds of identifier generation.

The two benefits offered by supporting compile time generation of
identifiers should theoretically be 1. efficiency and 2. ergonomics (via
the `widget_ids!` macro).

- In reality, 1. is redundant as conrod requires supporting dynamic
  index generation in order to support dynamic collections of widgets
  (i.e. lists, graphs, etc). Thus, all widgets require an index into the
  dynamically generated widget graph which means that our compile-time
  generated identifiers must be mapped to run-time generated identifiers
  either way, in turn losing the theoretical efficiency.
- 2. We should be able to adapt the current `widget_ids!` macro to
  support run-time generated identifiers.

In turn, it clearly seems worth removing support for compile-time
generation of identifiers in favour of focusing on improving the
ergonomics of run-time generation (the primary focus of this commit).

This commit removes the `widget::Index`, `widget::IndexSlot`,
`graph::NodeIndex` and `graph::index_map::IndexMap` types as well as the
`graph::index_map::GraphIndex` trait in favour of using `widget::Id`
exclusively.

Many parameters that used to require `I: Into<widget::Index>` now simply
take `widget::Id` directly.

The `widget_ids!` macro has been updated to generate a type that may be
used to produce a unique `widget::Id` for each given identifier. ~~Each
unique identifier acts as a `widget::Id` cache, similarly to the way
`IndexSlot` used to work.~~ *Edit: this is no longer the case as of [this commit](https://github.com/PistonDevelopers/conrod/pull/810/commits/e0f0373a76743c2522dd37b1bda1c5c023f4f796).*
The macro now also supports generating a growable `id::List` - a type that
abstracts much of the boilerplate around dynamically growable lists of
identifiers.

Pros:

- The process of generating identifiers is now the same when writing
  custom widgets as it is when using conrod from the top-level (i.e.
  just instantiating widgets and not writing custom ones). Previously it
  was common for top-level identifiers to be compile-time generated,
  while widget children identifiers would require run-time generation.
  Both now use run-time generation.
- Much of conrods internals are vastly simplified, removing a lot of
  confusing types and generic abstractions.
- New users will no longer be confused or repelled by the number of
  different identifiers.

Cons:

- In the examples, we can see that top-level identifier generation is
  now slightly more verbose than it previously was. I think there is
  still room for improving the ergonomics and flexibility of the updated
  `widget_ids!` macro. For example, the macro currently generates
  identifier fields as caches for `widget::Id`s, but provides no way to
  generate the `widget::Id` itself up front, despite this being the
  common use-case. This is currently the case because widgets do not
  have mutable access to the `Ui` (needed for the `new_unique_widget_id`
  method) when first instantiating the generated `Ids` type. We should
  be able to change this by providing mutable access to the `UiCell`
  within the `Widget::init_state` method. This would allow us to take
  the `Ui` as an argument to the `new` constructor method of the
  generated `Ids` type. I'm hoping to address this yet as a part of this PR.

Closes #791 

cc @tmerr


mitchmindtree:
Alright, I've added a commit that addresses the ergonomics issues mentioned in the parent comment.  I think this should be good to go now, though I'll leave this open for a little while longer to give anyone else a chance to add their thoughts :+1:

tmerr:
Wow, this makes it much easier to understand IDs. Great work!

mitchmindtree:
Alright going to merge. I've published these breaking changes in version **0.42.0**.

@tmerr Thanks for the feedback! I'm not sure I would have got around to this any time soon if you hadn't mentioned it, so thanks for bringing it up :+1:

PR #815, opened 2016-09-04
Fix floating point rounding error in List widget

mitchmindtree:
Fix published in **0.42.1**.


mitchmindtree:
Looks like travis failed due to a nightly-related bug in an upstream crate. Both stable and beta are still passing.

PR #817, opened 2016-09-04
Add support for attributes and fix warnings for struct generated by `widget_ids`.

mitchmindtree:
This slightly changes the syntax of the `widget_ids!` macro:

Before:
```rust
widget_ids! { Ids { button, toggles[] } }
```
After:
```rust
widget_ids! { struct Ids { button, toggles[] } }
```

This change to the syntax is in order to allow for the `Ids` struct to have attributes applied to it (including doc comments).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod"
-version = "0.42.1"
+version = "0.43.0"
 authors = [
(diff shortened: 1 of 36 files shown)


mitchmindtree:
Nightly is failing due to a bug in an old version of ndarray used by rusttype 0.2.0. Will hopefully be fixed by dylanede/rusttype#25. Going to merge as stable and beta are still passing.

Published as **0.43.0**.

PR #822, opened 2016-09-09
Added actions for Delete, Home and End to TextEdit widget

Boscop:
I added the following actions for Delete, Home and End to TextEdit widget:

Delete:
- no selection: delete char at cursor
- selection: delete selected text

Home:
- no selection: move cursor to beginning of line
- selection: go to beginning of the line where selection begins

End:
- no selection: move cursor to end of line
- selection: go to end of the line where selection ends

I'm not very familiar with the internals, I used the code from similar key actions and modified it.. Maybe the common code should be factored out into a function.

Other actions that I would still like to see:
- shift+left/right: select text by one char at a time
- ctrl+shift+left/right: select text by one word at a time (ctrl+left/right already moves cursor by one word)
- shift+home/end: select text until beginning/end of line
- ctrl+home/end: go to beginning/end of the whole text
- ctrl+shift+home/end: select text until beginning/end of whole text
- double click: select word
- triple click: select line / sentence / paragraph (in code editors it usually selects the line, in browsers the paragraph)

### DIFF
--- a/src/widget/text_edit.rs
+++ b/src/widget/text_edit.rs
@@ -471,6 +471,76 @@ impl<'a> Widget for TextEdit<'a> {
                             }
                         },
 
+                        // If `Cursor::Idx`, remove the `char` at the cursor.
+                        // If `Cursor::Selection`, remove the selected text.
+                        input::Key::Delete => {
+                            match cursor {
+
+                                Cursor::Idx(cursor_idx) => {
+                                    let idx_after_cursor = {
+                                        let line_infos = state.line_infos.iter().cloned();
+                                        text::glyph::index_after_cursor(line_infos, cursor_idx)
+                                    };
+                                    if let Some(idx) = idx_after_cursor {
+                                        let idx_to_remove = idx;
+
+                                        *text.to_mut() = text.chars().take(idx_to_remove)
+                                            .chain(text.chars().skip(idx_to_remove + 1))
+                                            .collect();
+
+                                        state.update(|state| {
+                                            let font = ui.fonts.get(font_id).unwrap();
+                                            let w = rect.w();
+                                            state.line_infos =
+                                                line_infos(&text, font, font_size, line_wrap, w)
+                                                    .collect();
+                                        });
+
+                                        let line_infos = state.line_infos.iter().cloned();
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Nice! Thanks for this

Did you test and check for bugs using the text_edit.rs example? I would do this myself but don't have any of these keys on my keyboard 😸

src/widget/text_edit.rs
```diff
@@ -575,6 +645,38 @@ impl<'a> Widget for TextEdit<'a> {
...
+                            }
+                        },
+
+                        input::Key::Home => { // move cursor to beginning.
+                            let mut line_infos = state.line_infos.iter().cloned();
+                            let line = match cursor {
+                                Cursor::Idx(idx) => idx.line,
+                                Cursor::Selection {start, ..} => start.line, // use first line of selection
+                            };
+                            match line_infos.nth(line) {
```
mitchmindtree: This match expr could be more concisely expressed as `if line_infos.nth(line).is_some() { ... }`

src/widget/text_edit.rs
```diff
@@ -575,6 +645,38 @@ impl<'a> Widget for TextEdit<'a> {
                             }
                         },
 
+                        input::Key::End => { // move cursor to end.
+                            let mut line_infos = state.line_infos.iter().cloned();
+                            let line = match cursor {
+                                Cursor::Idx(idx) => idx.line,
+                                Cursor::Selection {end, ..} => end.line, // use last line of selection
+                            };
+                            match line_infos.nth(line) {
```
mitchmindtree: As this match expr only has a single branch, we can use if let instead:

```rust
if let Some(line_info) = line_infos.nth(line) { ... }
```

Boscop:
Yes, I checked it in TextEdit and TextBox widgets.

Boscop:
Right, I copy/pasted code from other key actions and changed it so that it worked but I forgot to adapt the syntax. Did it now.

mitchmindtree:
Nice, looks good to me :+1: It would be nice to refactor out the common stuff between `Backspace` and `Delete`, but we can do this another time.

Issue #802, opened 2016-08-18
Continuous polling in Piston event loop, optimization


christolliday:
The core event loop used by Conrod seems to be a constant fixed rate loop where widgets are continuously polling for input events and continuously being updated, whether or not any events occur. This seems to be the cause of relatively high CPU usage even for a completely static GUI.

Am I misinterpreting the main loop, and if so, what accounts for the CPU usage of an idle Conrod app?

The best documentation for the Piston event loop I could find is here: https://github.com/PistonDevelopers/piston/issues/193

That type of event model makes sense for some games, where a continuously changing game environment is assumed, but not for a typical GUI app or even for many simpler games.

At any given point in time when there is no user input, no animations, no window resizing or any other events happening that would trigger UI changes, the CPU usage of a simple Conrod app should be zero or about to be zero. This would apply to all the current examples.

In addition to generally wasting resources, the current event model makes Conrod impractical for many use cases, especially for apps on battery constrained devices like smartphones or embedded devices.

At first glance, and at a very high level, I think moving to a more event driven design would involve:
- Breaking up the update method on widgets into two, one for receiving events and changing the state, and one for updating the widget. The update method would then only be called if an event took place which changed the widgets state.
- Replacing the continuous Piston event loop with something that only reacts to events, forwards them to widgets, and checks if updating and redrawing is necessary after any widget event callbacks are called
- For animations, allowing widgets to register for continuous, fixed rate event callbacks

Does this make sense and is it at all practical? Would such a change be consistent with Conrods immediate mode style?

mitchmindtree:
Thanks for the issue, definitely keen to optimise the idle CPU usage however we can.

> The core event loop used by Conrod seems to be a constant fixed rate loop where widgets are continuously polling for input events and continuously being updated, whether or not any events occur.

Conrod itself doesn't require any specific core event loop. It interacts with an application's loop at three distinct points:

- `Ui::handle_event` can be thought of as the input, where events are received, interpreted as `event::Ui`s and pushed to a queue.

- `Ui::set_widgets` is the stage at which all widgets are instantiated (updated). During this call, the events that have been queued via calls to `Ui::handle_event` are made available to each widget and can be accessed and filtered by widgets in their `Widget::update` method. The events are then cleared at the end of this stage when the `UiCell` is `drop`ped.

[...]

PR #831, opened 2016-09-15
Waiting event loop

christolliday:
This adds an optional waiting event loop that can be used with the `piston_window` backend, allowing apps to go completely idle when no events are being handled, saving background resources. See #802 

The new event loop takes a boolean parameter that indicates an animation is in progress, to trigger an update whether an event happens or not. By setting it to always true, the new event loop should have the same behaviour as the default piston event loop.

It works with piston_window but doesn't integrate perfectly with it, that is, it duplicates some code from piston_window, pulls in extra dependencies, and piston_window will still instantiate the old event loop and not use it.

The proper solution seems to be to either change the API of piston_window, which would be a breaking change, or include a forked version with conrod, since it seems like piston_window itself is not necessarily intended to be completely extensible. Not sure what the best thing to do is.
This works for now, but fixing up piston_window could remove some of the dependencies, although they are optional, and they will be pulled in by piston_window regardless.

Right now, it only works with `GlutinWindow`, until `wait_event` (blocking analog to `poll_event`) is added to the other piston back ends.

I modified some examples to use the new event loop, but there are some issues with the other ones, so I left those ones alone.
The issues seem to be caused is caused by code like this:

```
for _click in widget::Button::new()
    ...
    .label(&count.to_string())
    .set(ids.counter, ui)
{
    count += 1;
}
```

It seems that the widget layout is set, and then the click event is handled. This means that it takes one more update event before it displays the correct value.
This only manifests when you click the button and try not to move the mouse at all, because moving the mouse still triggers an update so it will update the widget anyway. 
As a side note, ideally events that don't cause any changes to state (most mouse move events) shouldn't be able to modify the UI.

I can hack around it by setting animation to true for one frame after the click happens, but I think the proper solution should be to make event handling and widget updating happening on the same frame, in one atomic event, so I left those examples alone for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,13 +49,25 @@ rusttype = "0.2.0"
 # - Rendering the `conrod::render::Primitives` yielded by `Ui::draw`.
 # Enables the `conrod::backend::piston_window` module.
 # Note: Use the `piston` feature for `piston_window` event conversions.
+#
+# `event_loop`
+# Enables use of piston_window backend with a non-polling event loop
 glium = { version = "0.15.0", optional = true }
 glutin = { version = "0.6.1", optional = true }
 piston2d-graphics = { version = "0.17", optional = true }
-piston_window = { version = "0.52", optional = true }
+piston_window = { version = "0.53", optional = true }
+pistoncore-glutin_window = { version = "0.30.0", optional = true }
+pistoncore-window = { version = "0.22.0", optional = true }
+gfx = { version = "0.12.0", optional = true }
+gfx_core = { version = "0.4.0", optional = true }
+gfx_device_gl = { version = "0.11.0", optional = true }
+
+
+
 [features]
-default = ["piston", "piston_window"]
+default = ["piston", "piston_window", "event_loop"]
 piston = ["piston2d-graphics"]
+event_loop = ["pistoncore-window", "pistoncore-glutin_window", "glutin", "gfx", "gfx_core", "gfx_device_gl"]
 
 [dev-dependencies]
 find_folder = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,23 +51,18 @@ rusttype = "0.2.0"
 # Note: Use the `piston` feature for `piston_window` event conversions.
 #
 # `event_loop`
-# Enables use of piston_window backend with a non-polling event loop
+# Enables use of piston_window backend with a non-polling event loop, only compatible with glutin_window
 glium = { version = "0.15.0", optional = true }
 glutin = { version = "0.6.1", optional = true }
 piston2d-graphics = { version = "0.17", optional = true }
-piston_window = { version = "0.53", optional = true }
+piston_window = { version = "0.54.0", optional = true }
 pistoncore-glutin_window = { version = "0.30.0", optional = true }
 pistoncore-window = { version = "0.22.0", optional = true }
-gfx = { version = "0.12.0", optional = true }
-gfx_core = { version = "0.4.0", optional = true }
-gfx_device_gl = { version = "0.11.0", optional = true }
-
-
 
 [features]
 default = ["piston", "piston_window", "event_loop"]
 piston = ["piston2d-graphics"]
-event_loop = ["pistoncore-window", "pistoncore-glutin_window", "glutin", "gfx", "gfx_core", "gfx_device_gl"]
+event_loop = ["pistoncore-window", "pistoncore-glutin_window", "glutin"]
 
 [dev-dependencies]
 find_folder = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ pistoncore-window = { version = "0.22.0", optional = true }
 [features]
 default = ["piston", "piston_window", "event_loop"]
 piston = ["piston2d-graphics"]
-event_loop = ["pistoncore-window", "pistoncore-glutin_window", "glutin"]
+event_loop = ["pistoncore-window", "pistoncore-glutin_window"]
 
 [dev-dependencies]
 find_folder = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,20 +49,15 @@ rusttype = "0.2.0"
 # - Rendering the `conrod::render::Primitives` yielded by `Ui::draw`.
 # Enables the `conrod::backend::piston_window` module.
 # Note: Use the `piston` feature for `piston_window` event conversions.
-#
-# `event_loop`
-# Enables use of piston_window backend with a non-polling event loop, only compatible with glutin_window
 glium = { version = "0.15.0", optional = true }
 glutin = { version = "0.6.1", optional = true }
 piston2d-graphics = { version = "0.17", optional = true }
-piston_window = { version = "0.54.0", optional = true }
-pistoncore-glutin_window = { version = "0.30.0", optional = true }
-pistoncore-window = { version = "0.22.0", optional = true }
+piston_window = { version = "0.55.0", optional = true }
+pistoncore-window = { version = "0.23.0", optional = true }
 
 [features]
-default = ["piston", "piston_window", "event_loop"]
-piston = ["piston2d-graphics"]
-event_loop = ["pistoncore-window", "pistoncore-glutin_window"]
+default = ["piston", "piston_window"]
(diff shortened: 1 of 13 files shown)


mitchmindtree:
Hey @christolliday, thanks for the PR and write-up! Sorry it took a while, been busy with a personal project over the past few days.

The new `event_loop` almost seems like a strictly more-flexible version of `piston_window`, in that it extends it with an option that provides slightly more efficiency for some use cases. I think I'd prefer if these changes could be applied upstream to `piston_window` itself if possible, rather than maintaining a fork as well as supporting the original.

> which would be a breaking change

How severe a departure is the new `event_loop` from the original? How much breakage would be involved?

> This means that it takes one more update event before it displays the correct value.

Yeah this is the case for that example, but isn't particularly an inherent limitation of conrod - more just a limitation of the `Button` in that particular use-case. If the user strictly requires that the label updates immediately, the user could instantiate the `Button` without the label, and then instantiate a `Label` widget immediately after, positioned at `.middle_of(button)`.

christolliday:
I agree about `piston_window`, I made a pull request here https://github.com/PistonDevelopers/piston_window/pull/164

As for the click events, that makes sense for that example, I started looking at the other ones though and I don't know if it will always work.
In the `image_button` example, clicking the button changes the colour of the background canvas, and the button is in front of the canvas. The z-order of drawing is controlled based on the order of instantiation right? Otherwise it could be possible to instantiate a new canvas after the Button but place it behind the Button. As is, I think it would have to take place over two frames?
Even if the Canvas can be drawn after the Button in that example, I can imagine there could be other examples where there is a circular dependency between the state of two or more widgets reacting to some event. It might not be possible to resolve it any other way than to call update all over again. Also in some cases it could be awkward to always arrange the order of instantiating widgets based on event flow.
Maybe just re-updating everything again is the best solution at the moment in some cases, I can at least add some helper code to make it easier to trigger a refresh.

christolliday:
Hi @mitchmindtree sorry for the long delay! I updated the PR based on the issue about `piston_window` and the issue I was mentioning in my last comment.
Basically I think you are right that re-arranging the order of the widgets is practical, the edge cases where that won't work are pretty rare, but do exist so an escape hatch is needed. Also, having to separate widgets into their dependent components is not as nice as being able to declare nested widgets.
I changed the API from taking an `animating` argument to having an `update` method, called either whenever an update is needed, or continuously for animation.
I used it to update the `counter` example with the new event loop, just to demo how it is used, if you think it's better to rearrange the widget order I could do that instead.

christolliday:
@mitchmindtree also sorry I didn't realize this was the cause of the broken Travis build before, saw it was in glutin_glium originally so I assumed it was something else.

mitchmindtree:
Hey @christolliday, thanks for waiting for me!

Rather than adding a new `event_loop` feature, I think I'd be fine with this going under the `piston` feature, as it seems like most cases that use `piston` or `piston_window` with conrod might want to use this? Otherwise, I think indicating that it's a piston-specific feature somehow would be nice - maybe `piston_wait_events` would be a better name?

Upon having a closer look, I can't help but feel this still seems like a bit of a hacky solution. From what I understand, the `PistonWindow` [still instantiates its own event loop internally](https://github.com/PistonDevelopers/piston_window/blob/master/src/lib.rs#L128), but then we create another separate event loop and pass that in via `next_event` (rather than `next`), so it ignores its inner `events` and uses the ones we pass it externally?

Ah, I just came across [this](https://github.com/PistonDevelopers/piston/issues/1112#issuecomment-248562616). Am I correct in thinking that, if this goes ahead, we won't need our own event loop at all? If that's a possiblity, I think I'd much prefer to wait. Adding a `wait_event` API to piston's `Window` trait sounds much nicer than adding our own entire loop module/feature. Do you mind if we wait and focus efforts there instead?

> Also, having to separate widgets into their dependent components is not as nice as being able to declare nested widgets.

Yeah I know what you mean. We can still nest the widgets involved in examples like `counter`, though this does require writing widgets for more specific use cases. For example, someone could write a `Counter` widget (or even just a `counter` function) which took an integer as an argument, instantiates a button, increments the integer if the button was pressed, instantiates a `Label` for the value and returns the new count as its `Widget::Event`. In actual code (non-example stuff) I think this would be the "proper" solution rather than calling something like `.update()`, as it avoids delaying the label update for a frame and updating the whole `Ui` for an extra frame.

That said, you're right in that there will still be cases where we'll need some way to continue updating and drawing (animated widgets being the prime example). Using `glutin` as an example, I imagine users could use `wait_events` while there is no animation occurring and switch to using `poll_events` for the duration of their widgets' animations, where some fixed-step loop is entered. It would be nice if the `Ui` had some way of tracking whether any of its `Widget`s are currently "updating"/"animating" so that it's easier to distinguish whether the event loop should be waiting or in some fixed-step mode.

christolliday:
Agreed this makes sense to put under the `piston` feature.

Also agree it is a little hacky with `piston_window` instantiating the other event loop. I proposed removing `WindowEvents` from that struct [here](https://github.com/PistonDevelopers/piston_window/pull/164) The problem is it would break/complicate the API for `piston_window`, and also seems to go against the goals of keeping piston_window very simple for basic use cases. I don't totally agree that `PistonWindow` shouldn't be made more flexible, but that seems to be the position that @bvssvni is taking. I was thinking in the future it might make more sense to add a forked version of piston_window, and it could be made less hacky then, but maybe we should deal with it now.
There are some ways I could fix it now without breaking the API for `PistonWindow` or making it more complicated/flexible in the default case, the only problem is, they would rely on having conrod library code in backend depend on `piston_window` which also goes against just having applications depend on piston_window.
The two ways are, use a trait object to store `WindowEvents` in `PistonWindow`, have an alternate constructor to `PistonWindow` to supply a custom event loop, and implement the `WindowEvents` trait for the custom event loop. My problem with that is, either every app would implement `WindowEvents` or conrod the library would, would breaks the application dependency rule. Also personally I prefer not having the event loop coupled to the window like it is in PistonWindow.

The other option which I prefer, would be to separate out and expose all the graphics related code in `piston_window`, maybe into a `GraphicsContext` struct? Then we could create a new `ConrodWindow` that shares GraphicsContext code with `PistonWindow`, but with an interface controlled by Conrod. That seems the most flexible, but without duplicating the bulk of `piston_window`. Still though, it results in library code being dependent on piston_window.

Considering all this, and https://github.com/PistonDevelopers/conrod/issues/835 maybe it does make sense to just include a fork of `piston_window`?

As for [this](https://github.com/PistonDevelopers/piston/issues/1112) unfortunately it doesn't get rid of the need for our own event loop, it just enables an event loop implementation to access a windows `wait_event` functionality via the common piston `Window` interface. Before, `wait_event` was only exposed by `GlutinWindow`, not the other backends. In our case it enables removing the dependency on `glutin_window`. It also just refines the API a bit and the `wait_event_timeout` method can be used to simplify the event loop added in this pull request, by moving the logic for waiting and timing out into glutin_window (or the other backends, which expose it directly). I can update this pull request to take advantage of that. 

I guess what you were alluding to is the idea that the default `piston/event_loop` could be updated to behave similar to this one and this wouldn't be needed. It's possible, but I still think there is a need to have different event loops based on the needs of a specific application. Having this in place also makes it much easier for conrod users to modify or plug in their own event loop without forking tons of code.

I can see what you mean about having a purpose built Counter widget, you could be right that that's the way to go eventually. I feel like there could be another way that doesn't require creating ad-hoc widgets, but that's another discussion I suppose.

For updating during animation, I agree the one thing missing is a way for the Ui to track whether any widgets are currently animating.

I created an example app that does this by using the `Widget::update` return value. The example defines a custom widget where `Widget::update` returns a flag that the widget is animating. The main loop, checks the return value for `widget.set(..)` and calls `events.update()` when the flag is set. 

It works for a simple example, but for a deeper hierarchy of widgets, the animation "Event" would need to get propagated all the way back up to the top to work which is not ideal.
Maybe it makes sense for each widget to store whether it is updating, maybe in the Widgets `CommonState`? Or just in the Ui directly. Then the main loop could check the Ui and pass that to the event loop.

phew that was a lot of text.. thanks for looking at this!

mitchmindtree:
@christolliday can we close this now that #850 is in?

christolliday:
the event loop here wasn't added by #850 but that being merged makes this code much simpler, I'll clean it up and open up a new one.

Issue #830, opened 2016-09-14
Continued discussion about piston-graphics


bvssvni:
See reddit discussion https://www.reddit.com/r/rust/comments/52jtl1/this_year_in_conrod_an_update_from_the_purerust/d7m9y8v

I'll move this discussion here so we get all information in one place.

bvssvni:
Some points of importance:

- Piston-Graphics can be used independently of whether you use a Piston window backend or the event loop. It can be used with an external window backend.
- The Piston-Graphics design is immediate mode, but allows a backend to override default methods for common primitives. By default it uses triangulation.
- Glium and Gfx backends stores colored triangles in a buffer to reduce draw calls. Problem is they can't take textures, so vector graphics alternating with text is a tricky combination.
- There is no software rasterize backend yet for Piston-Graphics. This might be better for some platforms.

bvssvni:
cc @mitchmindtree @kvark

kvark:
For the last point, I wonder if https://github.com/csherratt/rusterize could be revived. cc @csherratt

kvark:
> The Piston-Graphics design is immediate mode, but allows a backend to override default methods for common primitives. By default it uses triangulation.

Which default methods do you mean here?
Also, if the immediate mode is a problem, we might get a new API for building shapes. Basically, it would record all the buffer uploads triggered by compatible draw calls (as in, using the same shader and primitive type, blending, etc), and then return the buffer for reuse (on the following frames if UI layout doesn't change).

> Glium and Gfx backends store colored triangles in a buffer to reduce draw calls. Problem is they can't take textures, so vector graphics alternating with text is a tricky combination.

How is it currently solved without piston-graphics?

bvssvni:
@kvark Default trait methods https://docs.rs/piston2d-graphics/0.17.0/graphics/trait.Graphics.html

bvssvni:
Had no idea somebody was working on a 3D software rasterizer! Can this be made a Gfx backend?

kvark:
> Had no idea somebody was working on a 3D software rasterizer! Can this be made a Gfx backend?

It has been discussed and generally agreed upon, but this would be a non-trivial amount of work.

csherratt:
It could be revived, and it was fun to write. The performance was ok, but I would not call it fast. Bringing it to rust's stable branch might make the problem even worse since rust's stable branch does not support SIMD at the moment.

mitchmindtree:
[...]

PR #842, opened 2016-10-13
Add glutin_gfx example

pengowen123:
Adds an example equivalent to glutin_glium, but with gfx. I added the necessary dependencies to `[dev-dependencies]` in `Cargo.toml` because I didn't think a `gfx` feature would be worth adding without also adding a useful `backend/gfx.rs`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,5 +58,7 @@ default = ["piston", "piston_window"]
 piston = ["piston2d-graphics"]
 
 [dev-dependencies]
+gfx = "0.12.0"
+gfx_window_glutin = "0.12.0"
 find_folder = "0.3.0"
 rand = "0.3.13"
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Ooooo nice, thanks! I think you're right about holding off on the feature (I probably should have done the same with the glium one).

We may not need glium/gfx backends at all if we can land the optimisations necessary in piston-graphics (see [discussion here](https://github.com/PistonDevelopers/conrod/issues/830)), but I'd still like to add this for now just in case it doesn't work out.

Would you mind adding a line to `.travis.yml` that builds the example with the necessary feature like [this](https://github.com/PistonDevelopers/conrod/blob/master/.travis.yml#L23)?

pengowen123:
Okay, travis should now run the new example.

PR #844, opened 2016-10-20
Fix issue 827

clicketyclack:
Corner case generating white-on-white text + refined contrast calculation.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -123,7 +123,18 @@ impl Color {
     /// Return either black or white, depending which contrasts the Color the most. This will be
     /// useful for determining a readable color for text on any given background Color.
     pub fn plain_contrast(self) -> Color {
-        if self.luminance() > 0.5 { BLACK } else { WHITE }
+        match self {
+            Color::Hsla(h, s, l, _) => {
+                let (r, g, b) = hsl_to_rgb(h, s, l);
+                rgb(r, g, b).plain_contrast()
+            },
+            Color::Rgba(r, g, b, _) => {
+                let l = 0.2126 * r
+                      + 0.7152 * g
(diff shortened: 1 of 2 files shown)


src/color.rs
```diff
@@ -123,7 +123,18 @@ impl Color {
...
     /// useful for determining a readable color for text on any given background Color.
     pub fn plain_contrast(self) -> Color {
-        if self.luminance() > 0.5 { BLACK } else { WHITE }
+        match self {
+            Color::Hsla(h, s, l, _) => {
+                let (r, g, b) = hsl_to_rgb(h, s, l);
+                rgb(r, g, b).plain_contrast()
+            },
+            Color::Rgba(r, g, b, _) => {
+                let l = 0.2126 * r
```
mitchmindtree: Are these magic numbers for calculating luminance something you decided upon? Or are you referencing them from a standard somewhere?

Also, if we're going to use this expression to calculate luminance in this case, should we also be using it for the `luminance` method? Does this mean we end up with two different methods of determining `luminance` (luminosity channel from `Hsl` vs this expression)?

I do think what you've added is better for the `Rgba` case, though I'm wondering if a better fix would have been to change `fn luminance` to always convert to `Hsla` first and then use the luminosity channel.

mitchmindtree:
Hey thanks for the fix!

In the future, would you mind waiting for review before merging?

clicketyclack:
Yeah, sorry. New to this. Took the initiative to move forward based on this:
https://github.com/PistonDevelopers/rust-empty/issues/31 discussion on Jul 31 2015 and the linked issue https://github.com/PistonDevelopers/piston/issues/674

PR #847, opened 2016-10-23
Get rid of `PrimitiveType::Rectangle`

jonas-schievink:
Rectangles are now built using polygons, simplifying rendering.

### DIFF
--- a/examples/glutin_gfx.rs
+++ b/examples/glutin_gfx.rs
@@ -241,8 +241,6 @@ mod feature {
                 // Create vertices
                 while let Some(render::Primitive { id, kind, scizzor, rect }) = primitives.next() {
                     match kind {
-                        render::PrimitiveKind::Rectangle { color } => {
-                        },
                         render::PrimitiveKind::Polygon { color, points } => {
                         },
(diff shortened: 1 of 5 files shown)


mitchmindtree:
Hey thanks for the PR!

I did consider this originally, however as rectangles are so common in UIs, I thought it might be useful to include a `Rectangle` primitive in case a user wants to use instancing or some other shape-specific shader optimisation for them. I think I remember reading somewhere that Servo made some significant speed-ups by taking advantage of instancing for `Rectangle`s.

As an alternative to this, I've been thinking that it might be nice to offer an adaptor that converts the `Primitive`s into something simpler. E.g. we could have something like a `Primitives::triangles` method which renders shapes, lines and text as a series of triangles (though we'd probably still need a variant for the `Image`s I guess). Would be nice to grab your thoughts on this :)

jonas-schievink:
That makes a lot of sense! `Primitives::triangles` sounds good, too (but it won't be able to handle text). I guess this is basically the simplest the `Primitives` API can be, then, so I'll close this :)

PR #852, opened 2016-11-11
Complete `glium.rs` backend and update `glutin_glium.rs` example accordingly.

### LINKED ISSUE #752: Draw the remaining `render::Primitive`s within the `glutin_glium.rs` example
In #734 conrod introduced a more flexible method of rendering in a "backend agnostic" manner. Instead of requiring a constrained graphics backend, conrod now simply renders the entire `Ui` as a list of `render::Primitive`s.

#734 also introduced a `glutin_glium.rs` example in order to demonstrate this flexibility and demonstrate how to use conrod with glutin and glium directly. The PR correctly handles `Text`, however the other `Primitive`s are not yet handled.

- [x] `Rectangle`
- [x] `Lines`
- [x] `Polygon`
- [x] `Text`
- [ ] `Image` Could probably take inspiration from the way `Text` is rendered.

mitchmindtree:
The `glium` backend feature is now complete! Big thanks to @clicketyclack for kicking this off :)

The implementation uses a single shader program to render the entire GUI. For GUIs consisting of purely text and shapes, the entire GUI will be drawn with a single draw call. Otherwise a new unique draw call will be required for each new image and scizzor that appears, however we might be able to improve this using a TextureArray or mapping to a single texture in the future if necessary.

After updating the `glutin_glium.rs` example, I did some profiling and found the changes to be a significant improvement over the `all_widgets.rs` example (the same gui but rendered via piston_window).

The `all_widgets.rs` example (`piston_window` backend) idles at ~6.5% CPU, draws 60 frames of text at about ~60% CPU and at about ~65% CPU when scrolling.

On the other hand, the `glutin_glium.rs` example idles at about 1% CPU (though I anticipate this to drop further once we configure the gui to only update upon new input events). When using `ui.draw` instead of `ui.draw_if_changed` (and in turn uploading to the GPU roughly 60 times per second) the example peaks about about 7% - a very promising sign for heavily animated GUIs like timelines, etc. When scrolling continuously over large blocks of text it peaked at around 20%, though the vast majority of this time was spent in rusttype's glyph cache.

Here are some pics!

screen shot 2016-11-12 at 1 04 05 am

screen shot 2016-11-12 at 1 04 13 am

screen shot 2016-11-12 at 1 04 29 am

screen shot 2016-11-12 at 1 05 28 am

screen shot 2016-11-12 at 1 05 48 am

Room for improvement
---------------------------
- haven't had a chance to do line joins/caps yet, but that should be easy to add.
- might be able to improve performance slightly again by using an index buffer. By doing this, we'd probably be able to cut back on about a third of the vertices uploaded via the vertex buffer.
- use a TextureArray (or one big large texture) for uploading multiple images to the GPU at once, allowing to cut down on draw calls.

Closes #752.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,8 @@ piston = ["piston2d-graphics", "pistoncore-window", "pistoncore-event_loop", "gf
           "pistoncore-glutin_window", "piston-texture"]
 
 [dev-dependencies]
+find_folder = "0.3.0"
 gfx = "0.12.0"
 gfx_window_glutin = "0.12.0"
-find_folder = "0.3.0"
+image = "0.10.3"
 rand = "0.3.13"
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -48,9 +48,7 @@ fn main() {
 
     // Create our `conrod::image::Map` which describes each of our widget->image mappings.
     // In our case we only have one image, however the macro may be used to list multiple.
-    let image_map = image_map! {
-        (ids.rust_logo, load_rust_logo(&mut window.context)),
-    };
+    let image_map = support::image_map(&ids, load_rust_logo(&mut window.context));
 
     // Poll events from the window.
     while let Some(event) = events.next(&mut window) {
--- a/examples/glutin_gfx.rs
+++ b/examples/glutin_gfx.rs
@@ -301,7 +301,7 @@ mod feature {
                 let dpi_factor = dpi_factor as conrod::Scalar;
 
                 // Convert glutin event to conrod event, requires conrod to be built with the `glutin` feature
-                if let Some(event) = conrod::backend::glutin::convert(event.clone(), w, h, dpi_factor) {
+                if let Some(event) = conrod::backend::glutin::convert(event.clone(), &window) {
                     ui.handle_event(event);
                 }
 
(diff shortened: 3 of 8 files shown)


mitchmindtree:
Travis' failure is due to some breakage upstream when compiling gfx with nightly. Going to merge!

mitchmindtree:
@Boscop did you say you were using glium? If so you might be interested in this.

Boscop:
Thanks @mitchmindtree, great work!
Does this mean it's possible draw a HUD GUI over a glium-rendered scene now?

mitchmindtree:
@Boscop Yep!

Boscop:
Ok, great. If I want to draw my glium widgets in a certain rect area that will be spared out by conrod rendering, what's the best way to calculate the rect pos / size to align the areas?

mitchmindtree:
> rect area that will be spared out by conrod rendering

Hmmm not 100% sure what you mean by this?

Boscop:
I mean I want to render some of my GUI in conrod, and some of it in glium with shaders, so I draw the glium stuff first, and then conrod over it, but how can I prevent conrod from drawing its background over what I drew underneath?

Boscop:
@mitchmindtree For example, let's say I want to render my [midi view](https://i.imgur.com/oHylMER.png) with glium because it has so many rectangles for the notes and I want to batch draw them in a shader. But I want to draw the rest of [the GUI](https://i.imgur.com/IbTi3vo.png) with conrod so that I can have sliders and dropdown menus etc.
So I have to know the rect area where the midi view should be, so I can render it there, and tell conrod not to draw anything there, not even background. How should I do this?

Issue #854, opened 2016-11-12
Window resizing bug in piston backend


mitchmindtree:
Seems to have been introduced in #850.

Looks like this:

![screen shot 2016-11-12 at 6 40 36 pm](https://cloud.githubusercontent.com/assets/4587373/20236378/8b9ab996-a907-11e6-9fb9-70a8a2ed5d0b.png)

![screen shot 2016-11-12 at 6 40 45 pm](https://cloud.githubusercontent.com/assets/4587373/20236379/8f1b74c0-a907-11e6-97f0-839250385abc.png)

christolliday:
oh whoops, good catch, I actually saw this and assumed it was caused by some other change.
It looks like it's caused by not calling the correct `next` method for the window, I'm looking into a fix right now.

mitchmindtree:
Closed by #856

PR #855, opened 2016-11-12
Remove `event::Raw` in favour of using `event::Input`

### LINKED ISSUE #820: Use `piston::input::Input` as the `Raw` event type rather than `piston::input::Event`
Conrod only needs to know about the `Input` events, the other of the `piston::input::Event` variants (Render, AfterRender, Update, Idle) are not used.

I think the `Render` variant is currently used to get the initial window dimensions, however this is a little hacky and I think we'd be better off requiring that conrod is constructed with some initial window size and then monitors the `Input::Resize` event for changes to window size.

mitchmindtree:
From the commit message:

> Prior to this commit, conrod took `event::Raw` as the event type that
> drove forward the `Ui` and from which the `Ui` would generate
> higher-level `Event`s. It turns out we only ever really used the
> `event::Raw`'s `Input` variant, so to make our API more concise I
> thought we could remove the unnecessary indirection of `event::Raw`.
> 
> One problem this does raise is that on Mac OS, we use the `event::Raw`'s
> `Render` variant as a hack to track the window's size changes. This is
> only because glutin (winit) is unable to emit events during resizing on
> Mac OS. To get around the removal of this hack, I'm planning to submit a
> PR to glutin_window (used by the piston backend present in most of the
> examples) which will insert `Resize` events manually.
> 
> As a result of this change, we must also now pass the initial window
> dimensions to the `Ui`'s constructor.
> 
> This commit also updates the related event module docs.

Here is the PR that should fix the resize issues that appear on Mac OS when using the examples that depend on the piston feature: PistonDevelopers/glutin_window#99. Note that this is *not* related to the issue mentioned in #850 or #854.

Closes #820.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ pistoncore-event_loop = { version = "0.26.0", optional = true }
 piston2d-gfx_graphics = { version = "0.31.1", optional = true }
 piston-texture = { version = "0.5.0", optional = true }
 shader_version = { version = "0.2.0", optional = true }
-pistoncore-glutin_window = { version = "0.31.0", optional = true }
+pistoncore-glutin_window = { version = "0.32.0", optional = true }
 
 [features]
 default = ["piston"]
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -31,7 +31,9 @@ fn main() {
     let mut app = support::DemoApp::new();
 
     // construct our `Ui`.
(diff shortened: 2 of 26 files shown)


mitchmindtree:
Okydoke, the fix for `Resize` events was published in glutin_window v0.32.0, so I've updated our glutin_window dependency accordingly and resizing seems to work in the examples :+1:

PR #856, opened 2016-11-12
Fix window resizing bug

christolliday:
This is a fix for https://github.com/PistonDevelopers/conrod/issues/854

Turns out there was some uncaught ambiguity between the `next` method on `WindowEvents` and on the `EventWindow` trait implemented by `WindowEvents`. So that the EventWindow `next` method wasn't getting called, which let's the window handle resize events.
If you're interested in what happened it's basically this: https://play.rust-lang.org/?gist=eb607b5104023ee89bb8ca9f8b6f00a1&version=stable&backtrace=0. It's a little interesting/surprising that rust doesn't complain at all in this case. I wonder if this type of thing would get caught by clippy, maybe I'll check when the build starts working on nightly again.

Anyway it could be disambiguated by replacing `events.next(&mut window)` with `EventWindow::next(&mut events, &mut window)`.
But I thought it was better to just change the API to make the ambiguity impossible, so this changes it to `window.next(&mut events)` which is actually closer to the old `piston_window` API.

### DIFF
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -6,7 +6,7 @@ extern crate find_folder;
 mod support;
 
 use conrod::backend::piston::gfx::*;
-use conrod::backend::piston::{Window, UpdateEvent, OpenGL};
+use conrod::backend::piston::{Window, UpdateEvent, OpenGL, EventWindow};
 use conrod::backend::piston::core_events::{EventLoop, WindowEvents};
 use conrod::backend::piston::window as piston_window;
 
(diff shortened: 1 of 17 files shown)


mitchmindtree:
Woah nice catch! What a tricky bug. I reported the ambiguous method call issue at rust-lang/rust#37741 in case you're interested.

I think I prefer the original API of `events.next(&mut window)` (seems more sensical when reading), but if this is more practical for the API then it's probably fine :+1:

christolliday:
Awesome, thanks for making that issue, I'll be interested to see what comes of it for sure.

PR #857, opened 2016-11-13
Piston Backend Cleanup

mitchmindtree:
This makes some slight tweaks in the examples and API in an attempt to be a little more idiomatic. See commit messages for details. cc @christolliday

### DIFF
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -5,10 +5,9 @@ extern crate find_folder;
 
 mod support;
 
-use conrod::backend::piston::gfx::*;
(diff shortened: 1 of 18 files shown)


christolliday:
Looks good, the imports are looking a lot better.

Just to expand on why I added `core_events`, I was planning on adding a new `events` module with the waiting event loop from here https://github.com/PistonDevelopers/conrod/pull/831 once that's cleaned up. Then leaving the small `core_events` module just in case the old `pistoncore_event_loop` is needed for a few apps. Should have documented that a little better.

Anyway I found a bug in my unmerged event loop that might take a little while to fix, but when it's done I can have a look, I'll try to make the event loop swappable but not require any boilerplate in the common case, more like this style.

Issue #862, opened 2016-11-14
Image Button Alternate Image


tl8roy:
When a button is pressed/hover, it would be good to specify a different image.

The colour select is ok, but doesn't work that well for images.

tl8roy:
@mitchmindtree So I think I see see where you are going with the "Proper" solution for the Flat button colours. A generic ```Interaction``` emun sounds like a good way to me. Were you thinking of a Hashmap or something similar for storing them all in one place?

mitchmindtree:
Going to close this as this particular issue has been addressed in #927. Would be happy to discuss custom colours per interaction in another issue though :+1:

PR #864, opened 2016-11-14
Fix bugs in image `Button`

### LINKED ISSUE #861: Image Button Issues
With the Image Buttons, I am having a few issues.

Firstly, when a label is added, it is positioned under the image. Additionally, the image resizes to the width of the label.

Secondly, sometimes when the button is clicked, a dark background appears.
conrod image button

        // Button widget example button.
        if widget::Button::image(ids.img_dark_blue_button)
            .w_h(BUTTON_WIDTH_ARROW, BUTTON_HEIGHT_ARROW)
            //.mid_bottom_of(ids.background)
            .color(color::TRANSPARENT)
            //.floating(true)
            .border(0.0)
            .label("Testing")
            .label_color(TEXTWHITE)
            .label_font_size(32)
            //.image_color_with_feedback(color::RED)
            .set(ids.button, ui)
            .was_clicked()
        {
            println!("Yay!");

        }

mitchmindtree:
Fixes an issue where the image would draw after the label.

Fixes an issue where the image would inherit the size of the `Button`'s
label if there was one present.

Closes #861.

@tl8roy this should fix the problems you were having with label order and image size, however the behaviour of the mouse hover colouring is still the same. I like the idea you mentioned in #862, but that's probably best in another PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod"
-version = "0.46.0"
+version = "0.46.1"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/src/widget/button.rs
+++ b/src/widget/button.rs
@@ -10,6 +10,7 @@ use {
     Rect,
     Positionable,
     Scalar,
+    Sizeable,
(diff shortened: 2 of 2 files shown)


mitchmindtree:
Published in **0.46.1**.

Issue #870, opened 2016-11-16
Piston Backend: Render/Input event starvation when Update events consume too much CPU time


christolliday:
This is related to https://github.com/PistonDevelopers/conrod/issues/824

When running in debug mode, or on a slower machine, or when an app is complex enough that the update loop consumes 100% of the CPU, the `pistoncore-event_loop` continues posting `Update` events while the application starves for `Render` and `Input` events.
This happens in debug mode in any of the examples that have more than a few widgets on my 3 year old i7 (thinkpad t440s).

37979205a04735e7592610f2879773964db82750 removed `set_ups(60)` from the examples, which makes those examples use the default update rate of 120 updates per second (updates per second is different from frames per second in `pistoncore-event_loop`). This makes the issue appear more easily, but it happens with or without `set_ups(60)`.

mitchmindtree:
> removed set_ups(60) from the examples, which makes those examples use the default update rate of 120

Ahh my apologies, I removed the `set_ups` calls with the assumption that it would match the default fps.

@christolliday I'd love to get your thoughts on the `glutin_glium.rs` example. It makes me wonder whether this work adapting and maintaining the piston event loop and window abstractions is worth it when the `glutin` loop + polling seems to achieve what we're going for already in a simpler fashion with less indirection and more control in the hands of the user? By more control I mean:

- They can setup their own timed loops however they wish without the need for lots of trait implementations (as is the case in our piston/window backend)
- They can poll or wait for all queued input events at once
- They don't need to reach through a wrapper to access the window

I guess the more I think about it, the more I'm unsure of what benefits we are gaining by maintaining an implementation of the traits within the piston event and window libraries?

One thing that I do believe is useful to conrod is the conversion of different backend events to a common `Input` type, however as demonstrated with conrod's `glutin` feature this could be achieved with a single conversion function, rather than wrapping the entire window as is the case in `glutin_window`, `sdl2_window`, etc.

I'm curious to get your thoughts on all this.

christolliday:
Some good questions. I think most apps aren't going to need custom logic for polling/waiting, so I think it is useful to have some kind of event loop, separate from the app main loop that is reusable, while still allowing apps to easily use their own. I'm not sure how that can be done without some amount of indirection, unless we only support glutin, but I do agree with you about there being too much indirection in the piston backend.

[...]

Issue #824, opened 2016-09-09
Canvas example: tabs take 6 seconds to switch in debug mode


Boscop:
When running the canvas example in debug mode (on Windows 8.1), it takes about 6 seconds from clicking one of the tabs (FOO, BAR, BAZ) until the tab actually changes. (In release mode it's near instantaneous.)

christolliday:
This is solved for me by my pull request: https://github.com/PistonDevelopers/conrod/pull/831

This looks to be a performance issue caused by the matrix of buttons at the bottom of the canvas example taking a lot of work to update. On my machine, with the polling event loop, this example takes 100% CPU. It seems like the draw calls get delayed as it works to catch up with all the update calls in between. I profiled it and the biggest chunk of work is done by `matrix::Element::update`

Using a waiting event loop skips enough work for this to be responsive, but it still uses ~50% CPU in debug mode when moving the mouse around inside the window.

PR #871, opened 2016-11-16
Waiting event loop

christolliday:
This is a cleaned up version of https://github.com/PistonDevelopers/conrod/pull/831 now that it is easier to integrate with the new window.

From commits:
>Add alternative event loop based on waiting for events rather than continuous polling, so that it uses zero CPU while idle.
New event loop is built on top of new `wait_event` and `wait_event_timeout` methods added to `piston::window::Window`.
Event loop has `update` method to trigger animation at a fixed rate, or to refresh the view. Calling `update` each frame recreates the behaviour of the default piston event loop.

>In examples, prevent event loop from going idle for at least one frame after an input event.
This is necessary in the case when a widget does not update atomically in a single frame.
This enables the event loop to go idle when there is no input, while not making app implementation
more complicated.
Hopefully this is temporary until widgets either handle updates atomically, notify the event loop
that they require an update, or widgets are initialized after the events that determine the widget state,
(ie. if a button label is changed when the button is pressed, the button label is created after
the button, and set based on whether the button was pressed.)

I originally intended to leave `pistoncore-event_loop` optionally supported by the `core_events` module. But with this change to the examples, (made easier by the recent change from `Raw` events to `Input` events) all apps should be able to use this event loop without any major changes.

It's still not 100% efficient, because it runs occasional redundant updates, however it's a major improvement over constantly running them 60+ times per second!

Also this event loop prevents the input/render event starvation issue here https://github.com/PistonDevelopers/conrod/issues/870 and resolves this issue https://github.com/PistonDevelopers/conrod/issues/824

There's one outstanding problem, the close window button doesn't work because of this: https://github.com/PistonDevelopers/glutin_window/issues/100.
Considering that, it might make sense to wait to merge until that issue is resolved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ path = "./src/lib.rs"
 [dependencies]
 daggy = "0.4.0"
 num = "0.1.30"
-pistoncore-input = "0.14.0"
+pistoncore-input = "0.15.0"
 rusttype = "0.2.0"
 
 # Optional dependencies and features
@@ -55,16 +55,15 @@ piston2d-graphics = { version = "0.19", optional = true }
 gfx = { version = "0.12.0", optional = true }
 gfx_core = { version = "0.4.0", optional = true }
 gfx_device_gl = { version = "0.11.0", optional = true }
-pistoncore-window = { version = "0.23.0", optional = true }
-pistoncore-event_loop = { version = "0.26.0", optional = true }
+pistoncore-window = { version = "0.24.0", optional = true }
 piston2d-gfx_graphics = { version = "0.33.1", optional = true }
 piston-texture = { version = "0.5.0", optional = true }
 shader_version = { version = "0.2.0", optional = true }
-pistoncore-glutin_window = { version = "0.32.0", optional = true }
+pistoncore-glutin_window = { version = "0.33.0", optional = true }
 
 [features]
 default = ["piston"]
-piston = ["piston2d-graphics", "pistoncore-window", "pistoncore-event_loop", "gfx",
+piston = ["piston2d-graphics", "pistoncore-window", "gfx",
           "gfx_core", "gfx_device_gl", "piston2d-gfx_graphics", "shader_version",
           "pistoncore-glutin_window", "piston-texture"]
 
--- a/examples/all_widgets.rs
+++ b/examples/all_widgets.rs
@@ -56,6 +56,7 @@ fn main() {
         // Convert the piston event to a conrod event.
         if let Some(e) = piston::window::convert_event(event.clone(), &window) {
             ui.handle_event(e);
+            events.update();
         }
 
         event.update(|_| {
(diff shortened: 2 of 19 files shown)


christolliday:
New commits fix the close window issue, was fixed in `glutin_window`.
Nightly seems to be broken by https://github.com/rust-lang/cargo/issues/3340

christolliday:
Thanks for having a look at this! No worries on the wait, it's not holding me back from anything, I can understand your concerns, I just think that for a new user, running a complicated example (see `canvas`) and starting htop can be a little alarming.

I can see why you might not want it part of the piston feature, so maybe it should be part of the examples, the only problem with that is it won't be updated in apps if they update conrod versions, but I guess things are still really early so things won't be stable between conrod updates anyway.

That said, I disagree with the idea that apps that need something like this are the exception and that it implies we would need to provide alternate loops for other use cases. The typical use case for a GUI library is still an event driven single window, desktop app, as all the examples are. Making that use case easy is still useful even if you are doing something more involved than that.

Your example loop looks mostly good, and it does show the parts of this loop that are vestiges of the piston loop. Specifically, using RenderArgs, and splitting up Update/Render/AfterRender into different steps. Removing that I could probably get rid of the 'State' enum and make this loop closer to your example.

Aside from that, the main benefit of this loop over your function is that it can still update the UI at a fixed rate when needed, and it caps the minimum update rate at 60fps, so that if a series of events all happen at once, it doesn't update and render faster than that, while also enabling a continuous, 60fps animation if updates are requested by the UI. So it is still compatible with realtime games or media, or just a typical UI that occasionally shows an animation. With your code, animations need to manage time manually rather than being able to assume each update represents a frame. This could be good or bad, but I think animations are easier to implement this way.
By polling and collecting events before updating, your loop does prevent the worst issues of the old (current) piston loop, that is, the UI becoming unresponsive, but I'm pretty sure it will still spike the CPU and update at the highest frame rate your CPU can run it at for as long as there are a series of input events coming in at that rate, a mouse drag, for example.
I think providing something that solves these problems makes it so the average user has one less thing to worry about.

Anyway if you add that and separate your code into a function, it will be dependent on some backend. The function would need to call poll/wait event on a window inside that function, so it would take a window, and update flag, and return an input event or a signal to render/close. Either that or pass it callbacks to call your handle_event, and render functions, and/or abstract over the window one more time.

So having a struct containing a window reference and update flag isn't too far removed from just a function that takes those arguments, I think it provides slightly better separation of concerns than a function but it's essentially the same.

christolliday:
sorry for the so much text btw! you can probably just read the second half.

Anyway would it help if I simplify the state part, and put in examples/support/piston_events?

christolliday:
the new support event loop resolves all of the issues this set out to resolve, so this is no longer needed. I think the only functional difference is it delivers events as a batch every frame rather than handling them as they come in, but I can't think of a real reason why it shouldn't. Also it can't change the fps, but that's probably not needed either.

PR #872, opened 2016-11-16
Fix `TextEdit` height inference when unrestricted. Auto-scroll text when cursor would exceed scrollable area due to input.

### LINKED ISSUE #866: Scrollable TextEdit
Is it somehow possible to make a `TextEdit` scrollable?

mitchmindtree:
Currently `TextEdit` uses the default height inference when no height is specified, which is to take the height of the previously instantiated widget.

This PR fixes this behaviour so that if the `TextEdit`'s `restrict_to_height` field is set to `false`, its height will be inferred as the total height of the styled, line-wrapped text that it displays.

The `text_edit.rs` example has been updated to demonstrate this behaviour.

One problem that was highlighted while testing this is that the the parent container does not automatically scroll down when text is entered that would cause the size of the text to become greater than the size of the scrollable parent. This could possibly be fixed in a future PR by having the `TextEdit` "bubble up" a scroll event to its immediate y-axis-scrollable parent via the `Ui::scroll_widget` method. I'd add this now but it's midnight and I'm ready to 😴 

In the mean-time, a user may be able to emulate this behaviour by tracking the `TextEdit`'s height, comparing it to its parent's height and creating scroll events via the `Ui::scroll_widget` method as necessary themselves. Edit: although I guess what we really need is for the cursor position to be the thing that affects scrolling, not the total text height. Will look into this next time I get a chance.

Closes #866.

@Jascha-N hopefully this helps! I'll publish the fix to crates.io when travis passes :+1:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod"
-version = "0.46.1"
+version = "0.46.2"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
--- a/examples/text_edit.rs
+++ b/examples/text_edit.rs
@@ -5,7 +5,7 @@ use conrod::backend::piston::{self, Window, WindowEvents, OpenGL};
 use conrod::backend::piston::event::UpdateEvent;
 
 widget_ids! { 
-    struct Ids { canvas, text_edit }
+    struct Ids { canvas, text_edit, scrollbar }
 }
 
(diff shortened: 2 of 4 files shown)


mitchmindtree:
Couldn't help myself, ended up adding the auto-scrolling on cursor input too. For some reason hitting the `return` key doesn't always scroll the parent properly (will look into this another time), but the rest of the cursor-adjusting input (including mouse) seems to auto-scroll fine.

Travis passes on stable and beta, only nightly fails.

Published as **0.46.2**.

Issue #884, opened 2016-11-25
conrod with glutin/glium backend does not work with OpenGL ES


TimBednarzyk:
There seems to be at least 2 issues causing this. The first is that the shader code for the glium backend does not have a valid version for OpenGL ES (Should be `#version 330` for OpenGL ES 3.0, or with some other changes to the shader code `#version 100` for OpenGL ES 2.0). I made a local copy and changed the shader code accordingly. However, I still get another error:
```
thread 'main' panicked at 'assertion failed: ctxt.version >= &Version(Api::Gl, 3, 2) ||
    ctxt.extensions.gl_arb_sampler_objects', /home/kuro-kun/.cargo/registry/src/github.com-1ecc6299db9ec823/glium-0.15.0/src/sampler_object.rs:22
stack backtrace:
   1:     0x564a7220cdb9 - std::sys::backtrace::tracing::imp::write::hd4b54a4a2078cb15
   2:     0x564a722108dc - std::panicking::default_hook::_{{closure}}::h51a5ee7ba6a9fcef
   3:     0x564a7220fb89 - std::panicking::default_hook::hf823fce261e27590
   4:     0x564a72210218 - std::panicking::rust_panic_with_hook::h8d486474663979b9
   5:     0x564a71f9e933 - std::panicking::begin_panic::h94e6fb50d7864116
                        at /buildslave/rust-buildbot/slave/stable-dist-rustc-linux/build/obj/../src/libstd/panicking.rs:384
   6:     0x564a7206bb73 - glium::sampler_object::SamplerObject::new::h7657da2b8588ab98
                        at /home/kuro-kun/.cargo/registry/src/github.com-1ecc6299db9ec823/glium-0.15.0/src/sampler_object.rs:3
   7:     0x564a7206c026 - glium::sampler_object::get_sampler::hc16e650b4458f717
                        at /home/kuro-kun/.cargo/registry/src/github.com-1ecc6299db9ec823/glium-0.15.0/src/sampler_object.rs:105
   8:     0x564a71db527f - glium::uniforms::bind::bind_texture_uniform::h3040645f16479981
                        at /home/kuro-kun/.cargo/registry/src/github.com-1ecc6299db9ec823/glium-0.15.0/src/uniforms/bind.rs:588
   9:     0x564a71db3b55 - glium::uniforms::bind::bind_uniform::h08b759a7b6720359
[...]

PR #885, opened 2016-11-27
Add support for OpenGL ES 3.0+

TimBednarzyk:
Fixes the first issue in #884 by adding shaders compatible with OpenGL ES 3.0+.


mitchmindtree:
Awesome! Thanks a lot for this @TimBednarzyk :)

PR #882, opened 2016-11-23
Improve EnvelopeEditor UX by inserting and removing points on Press rather than Release

mitchmindtree:
This also closes #880 cc @Kingsquee

PR #886, opened 2016-11-28
Publish 0.47.0 - improved EnvelopeEditor UX and glium GL ES 3+ support

mitchmindtree:
Bumped the minor version in this case due to very slight breaking change
in EnvelopeEditor that changes the behaviour of point insertion and
removal slightly. See #882 for details.

Includes the patch for supporting OpenGL ES 3.0+ via an alternative
shader program in the glium backend feature. See #885.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod"
-version = "0.46.2"
+version = "0.47.0"
 authors = [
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Travis is still failing on nightly due to a gfx error, but still builds on beta and stable. Published and merging.

PR #891, opened 2016-12-13
Validate Cursor in TextEdit widget. Clamp to text bounds if out of range.

mitchmindtree:
This adds a method to the `text::cursor::Index` for clamping to some given text described via an iterator yielding `text::line::Info`s.

Use this new method in the `TextEdit::update` in order to validate the position of the `Cursor`.


mitchmindtree:
Published as **0.47.2**.

PR #897, opened 2016-12-24
Fix List Scrollbar visibility following first widget update

mitchmindtree:
When built with `.scrollbar_next_to()`, the List would never instantiate
its Scrollbar within the List's first `Widget::update`, causing the List
to slightly jitter to the left as the Scrollbar would appear a frame or
two later. This commit fixes this behaviour.


mitchmindtree:
Published as **0.47.4**.

Issue #895, opened 2016-12-22
Reusing a texture in an image map


TimBednarzyk:
I'm trying to use one texture for multiple images. My texture consists of all of a bunch of icons packed together in order to save on the number of textures. However, with the glutin/glium backend, I can't seem to have an image map of references to `glium::texture::Texture2d`s like so:
```
let image_map = image_map! {
    (ids.image1, &icons),
    (ids.image2, &icons)
};
```
The error I get when I use references is this:
```
error[E0271]: type mismatch resolving `<&conrod::glium::Texture2d as std::ops::Deref>::Target == conrod::glium::texture::TextureAny`
   --> src/main.rs:150:16
    |
150 |       renderer.fill(&display, primitives, &image_map);
    |                ^^^^ expected struct `conrod::glium::Texture2d`, found struct `conrod::glium::texture::TextureAny`
    |
    = note: expected type `conrod::glium::Texture2d`
    = note:    found type `conrod::glium::texture::TextureAny`
    = note: required because of the requirements on the impl of `conrod::backend::glium::TextureDimensions` for `&conrod::glium::Texture2d`

error[E0277]: the trait bound `for<'a> conrod::glium::uniforms::Sampler<'a, &conrod::glium::Texture2d>: conrod::glium::uniforms::AsUniformValue` is not satisfied
   --> src/main.rs:154:16
    |
154 |       renderer.draw(&display, &mut target, &image_map).unwrap();
    |                ^^^^
    |
    = help: the following implementations were found:
    = help:   <conrod::glium::uniforms::Sampler<'a, conrod::glium::texture::Texture1d> as conrod::glium::uniforms::AsUniformValue>
    = help:   <conrod::glium::uniforms::Sampler<'a, conrod::glium::texture::CompressedTexture1d> as conrod::glium::uniforms::AsUniformValue>
    = help:   <conrod::glium::uniforms::Sampler<'a, conrod::glium::texture::SrgbTexture1d> as conrod::glium::uniforms::AsUniformValue>
    = help:   <conrod::glium::uniforms::Sampler<'a, conrod::glium::texture::CompressedSrgbTexture1d> as conrod::glium::uniforms::AsUniformValue>
    = help: and 55 others

```
Is there any way I can use one texture for multiple images?

tl8roy:
If I was going to hazard a guess, 
http://docs.piston.rs/conrod/conrod/widget/primitive/image/struct.Image.html#method.source_rectangle is probably the method to use.

[...]

PR #906, opened 2017-01-19
Add Alt Button Images and Custom Colours for Hover, Clicked

tl8roy:
This is my first pull request ever so hopefully this goes well.

I have implemented my request in #862. The Image Button constructor now has a second option parameter that takes an image. When the button is pressed, this image is shown instead of the other one.
I also added another image in assets to use.

I also added the ability to specify the colours shown when the button is hovered over or clicked. This defaults to Transparent.

I have tried to follow the style as much as I can, but there is certainly some items that could be re-factored.


tl8roy:
Probably should also add that I have successfully used the new features on the project code base.

src/widget/button.rs
```diff
@@ -128,11 +136,12 @@ impl Iterator for TimesClicked {
 impl<'a> Button<'a, Image> {
 
     /// Begin building a button displaying the given `Image` on top.
-    pub fn image(image_id: widget::Id) -> Self {
+    pub fn image(image_id: widget::Id, alt_image_id: Option<widget::Id>) -> Self {
```
mitchmindtree: Rather than adding an `Option` argument to the `image` constructor, I think this might be more idiomatic as a new builder method, so that it could be called like this:

```rust
Button::image(image_id).click_image(alt_image_id).set(id, ui);
```

Edit: on second thoughts, perhaps `.press_image` makes more sense than `.click_image`.

src/widget/button.rs
```diff
@@ -77,6 +83,8 @@ pub struct Image {
     pub color: ImageColor,
     /// The rectangular area of the original source image that should be displayed.
     pub src_rect: Option<Rect>,
+    /// The id of an Alternate image to show when pressed
+    pub alt_id: Option<widget::Id>,
```
mitchmindtree: Perhaps a better name field for this would be `press_id`, to clarify that it is the image used while the button is pressed.

src/widget/button.rs
```diff
@@ -247,18 +258,29 @@ impl<'a, S> Widget for Button<'a, S>
...
+            let mut is_down = false;
             let color = input.mouse().map_or(color, |mouse| {
                 if mouse.buttons.left().is_down() {
-                    color.clicked()
+                    is_down = true;
+                    if let Some(clicked_color_destruct) = style.clicked_color {
+                        clicked_color_destruct
+                    } else {
+                        color.clicked()
+                    }
```
mitchmindtree: When using an `if let` with an `else` branch, I normally opt for using `match` instead as it tends to be a little more concise for multiple branches:
```rust
match style.clicked_color {
    Some(clicked_color_destruct) => clicked_color_destruct,
    None => color.clicked(),
}
```
This is just personal preference, but just thought I'd mention it!

src/widget/button.rs
```diff
@@ -213,6 +222,8 @@ impl<'a, S> Button<'a, S> {
 
     builder_methods!{
         pub enabled { enabled = bool }
+        pub hover_color { style.hover_color = Option<Color> }
+        pub clicked_color { style.clicked_color = Option<Color> }
```
mitchmindtree: I don't think the `hover_color` or `clicked_color` methods should take an `Option` argument, as they are already optional to call. Would you mind changing these to take a `Color` directly?

src/widget/button.rs
```diff
@@ -348,6 +370,15 @@ impl Show for Image {
...
             ImageColor::None => None,
         };
-        image.set(id, ui);
+
+        let mut image_id = id;
+        if let Some(alt_image) = alt_id {
+            if is_down {
+                image_id = alt_image;
+            }
+        }
```
mitchmindtree: These 6 lines could be expressed using a match expression:
```rust
let image_id = match alt_id {
    Some(alt_id) if is_down => alt_id,
    None => id,
};

mitchmindtree:
I'm currently working on #895, just letting you know that I'll likely address that first and then rebase your PR here on top of those changes :+1:

tl8roy:
So I tried to update my fork and managed to screw it up.
What is the best process to merge changes when a large portion of the code changes?

mitchmindtree:
> What is the best process to merge changes when a large portion of the code changes?

Normally I find `git rebase` is a better option than `merge` for these cases, as it allows you to re-construct your commits as though they originally branched from the branch that you are rebasing upon. E.g. if you were working on branch `button_images`, you could do `git rebase master` and follow the prompts to update your commits as though they originally branched from the current head of `master`. This avoids adding lots of messy `merge` commits which often aren't very useful when looking back in history. After rebasing, you'll often be required to force a `git push` as `rebase` does modify history (although only for the commits on your branch so it is not a big deal).

It looks like some indenting is broken and that there are leftover `.orig` files in there. It might be easier if I have a go at implementing this now :+1: I'll do a PR and ping you shortly.

tl8roy:
Yeah, sorry about that.

mitchmindtree:
Closing in favour of #927.

PR #915, opened 2017-01-28
Use Condvar to make the example EventLoop thread-safe

pedrocr:
The event loop used to need to be mutable and was thus hard to use when you had several threads that needed to call needs_update(). Use std::sync::Condvar and friends to make it so the EventLoop doesn't need to be mutable at all and handles several threads properly.

next() will now busy loop at 60FPS while waiting for input. Ideally there would be a way to have the glium Window also fire needs_update() calls and then we could bump the timeout to something much higher or even just wait forever until there's an event. I don't see anything in the glium API that lets us do that cleanly though and this busy waiting is cheap (but may be a waste of power on mobile/laptops).

### DIFF
--- a/examples/support/mod.rs
+++ b/examples/support/mod.rs
@@ -15,6 +15,8 @@ extern crate rand;
 
 use conrod;
 use std;
+use std::sync::{Arc, Mutex, Condvar};
+use std::time::Duration;
 
 #[cfg(feature="glium")] use conrod::backend::glium::glium;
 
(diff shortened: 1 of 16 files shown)


examples/support/mod.rs
```diff
@@ -364,43 +366,41 @@ pub fn gui(ui: &mut conrod::UiCell, ids: &Ids, app: &mut DemoApp) {
...
+    pub fn next(&self, display: &glium::Display) -> Vec<glium::glutin::Event> {
         let mut events = Vec::new();
-        events.extend(display.poll_events());
 
-        // If there are no events and the `Ui` does not need updating, wait for the next event.
-        if events.is_empty() && !self.ui_needs_update {
-            events.extend(display.wait_events().next());
+        // FIXME: This will busy loop at 60FPS, ideally there would be a way to have the glium
+        //        display also fire needs_update() calls and then we could bump the timeout
+        //        to something much higher or even just use wait()
```
mitchmindtree: I think we might only get benefits from `Condvar` if the conrod loop is run on a separate thread from the `winit` events loop, which is certainly possible though a little more involved and probably a bit complicated for the examples (which this simple event loop implementation is for). I don't think there's any way for `winit` events to `notify` its own thread as they are waited for on the same thread that's being `wait`ed by the `Condvar`.

pedrocr:
See #918 for how I implemented a solution for this that's really simple.

PR #913, opened 2017-01-28
Publish new version

mitchmindtree:
Breaking changes include:

- Introduction of new `image::Id` type for referring to images #912 
- Remove glutin feature in favour of new more flexible winit feature #910 
- Simplification of piston feature #909

PR #917, opened 2017-01-29
Add a slimmer `Input` enum with only the variants that are relevant to conrod.

mitchmindtree:
This should simplify conrod's `Input` event API, making it clear which
events are necessary to drive the `Ui` forward.

This also fixes an issue appearing in downstream crates where conrod
loops that are run on a separate thread were broken due to `Input` no
longer being `Send` or `Sync`. See [here](https://github.com/PistonDevelopers/piston/pull/1149#pullrequestreview-18985967) for details.

Edit: Also, the `Motion` enum re-export was moved to the `input` module along with the rest of the remaining re-exported `pistoncore-input` types.

### DIFF
--- a/src/backend/piston/event.rs
+++ b/src/backend/piston/event.rs
@@ -1,7 +1,8 @@
 //! A backend for converting piston events to conrod's `Input` type.
 
 use {Point, Scalar};
-use event::{Input, Motion};
+use event::Input;
+use input;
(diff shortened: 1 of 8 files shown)


mitchmindtree:
To clarify, this basically returns the `Input` enum back to what it was before updating to the latest version of pistoncore_input in #913.

PR #918, opened 2017-01-29
Implement a threaded version of the `all_winit_glium.rs` example

mitchmindtree:
cc @pedrocr @Hyperchaotic you might be interested in this.

### DIFF
--- a/examples/all_winit_glium_threaded.rs
+++ b/examples/all_winit_glium_threaded.rs
@@ -1,0 +1,195 @@
+//! This example behaves the same as the `all_winit_glium` example while demonstrating how to run
+//! the `conrod` loop on a separate thread.
+
+#[cfg(all(feature="winit", feature="glium"))] #[macro_use] extern crate conrod;
+#[cfg(all(feature="winit", feature="glium"))] mod support;
+
+fn main() {
+    feature::main();
+}
+
+#[cfg(all(feature="winit", feature="glium"))]
+mod feature {
+    extern crate find_folder;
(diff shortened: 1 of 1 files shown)


mitchmindtree:
@pedrocr, @Hyperchaotic sorry about the short message before, I had to leave in a hurry!

To clarify, this demonstrates a threaded conrod app without the need for `Condvar` or a busy loop. We do this by running the conrod loop on a separate thread and signalling the `winit::Window` to wake up when it's time to render using a `winit::WindowProxy` instead. Note that the `event_{tx/rx}` channel could be changed to send any kind of message to the conrod thread. E.g. conrod could receive updates from some background thread as well as events from the main `winit` by changing the channel to send something like this:
```rust
enum Message {
    Event(conrod::event::Input),
    SomeOtherProcess(SomeUpdate),
}
```

pedrocr:
I found an even better way of doing this that after opening a bug report in glutin about the events API:

https://github.com/pedrocr/chimper/blob/1cdfdab0fbbff82f88826773687effb72432137c/src/event.rs

Since glutin has the WindowProxy struct that can safely be passed to another thread all you have to do is use a blocking wait_events() for the main loop and then call wakeup_event_loop() on the proxy from the other thread when you have something else you want to react to. Now my app doesn't wake up 60 times a second even when there's nothing to do.

This will do more than 60fps if there are a lot of events but that can easily be worked around by saving the frame times in the struct (making it mut again) and having the WindowProxy struct be another non-mutable one.

Issue #919, opened 2017-01-29
conrod::image::Map needs remove()/replace()


pedrocr:
I'm writing an image viewer with conrod so I create new images by reading them from the disk. Right now whenever I switch image I just use insert() to get a new conrod::image::Id and pass that on to widget::Image::new(). This works fine but I'm now leaking images. It would make sense to be able to have insert() and replace() methods so I can more easily keep the image map clean of old images I've since stopped using.

mitchmindtree:
The only thing that stopped me from adding a `remove(id)` method originally was that the user may invalidate their `image::Id`s, and I was wondering if there might be some way we could avoid this... In reality, I think all we can do is either `panic!` if a user attempts to use an invalid `Id` with a useful message or simply not draw the image. Users will just have to be careful to not use `image::Id`s after their associated image data has been removed.

I'd be happy to accept a PR that added both of these methods, otherwise I'll try to get around to it myself soon.

tl8roy:
Perhaps the image widget should return a result when ```set()``` is finally called?
That way the programmer can deal with it in the manner that they want to.

mitchmindtree:
@tl8roy the `image::Id` isn't actually used by the `Image` widget during its instantiation so I don't think returning `Result`s here would help. The problem arises when the `Id`s are used to index into the map which doesn't occur until rendering.

Closed by #923.

PR #923, opened 2017-01-31
Add replace()/remove() to image::Map

pedrocr:
As discussed in #919 It is often useful to be able to remove an image (to not leak) or to replace it (to not have to continuously keep track of new Id's when swapping images). Implement these in the natural way using the HashMap.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -72,13 +72,7 @@ impl<Img> Map<Img> {
         self.map.get_mut(&id)
     }
 
-    /// Inserts the given widget-image pair into the map.
-    ///
-    /// If the map did not already have an image associated with this widget, `None` is returned.
-    ///
-    /// If the map did already have an image associated with this widget, the old value is removed
-    /// from the map and returned.
-    ///
+    /// Inserts the given image pair into the map.
     /// Note: Calling this will trigger a redraw the next time `Ui::draw_if_changed` is called.
     pub fn insert(&mut self, img: Img) -> Id {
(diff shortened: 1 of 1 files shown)


src/image.rs
```diff
@@ -72,13 +72,7 @@ impl<Img> Map<Img> {
...
     }
 
-    /// Inserts the given widget-image pair into the map.
-    ///
-    /// If the map did not already have an image associated with this widget, `None` is returned.
-    ///
-    /// If the map did already have an image associated with this widget, the old value is removed
-    /// from the map and returned.
-    ///
+    /// Inserts the given image pair into the map.
```
mitchmindtree: Ahh thanks for removing the old comment here!

Would you mind adding that this method also returns an `Id` associated with the added image, and that the user must store this in order to use, modify or remove the inserted image?

src/image.rs
```diff
@@ -89,6 +83,20 @@ impl<Img> Map<Img> {
         id
     }
 
+    /// Replaces the given image in the map if it exists. Returns the image or None.
+    /// Note: Calling this will trigger a redraw the next time `Ui::draw_if_changed` is called.
```
mitchmindtree: Would you mind adding another line break before the `Note:` lines on this method, the `insert` method and the `remove` method docs? It's just that `rustdoc` normally doesn't add a line break when rendering unless there's at least one blank line in between.

mitchmindtree:
Going to merge this and make these changes myself so that I can publish the next version. Thanks!

PR #926, opened 2017-01-31
glium backend: Do not do gamma conversion on images

### LINKED ISSUE #914: `glium` backend: The `render::Primitive::Image` variant should not do gamma conversion and instead rely on the user selecting the appropriate glium `Texture` type for their image.
It seems conrod is doing it's own sRGB gamma conversion when that should be the job of glium. Ideally that would be left to glium so you can do things like setup f32x3 RGB textures and let glium handle the displaying to the screen in the way that's best for the particular setup (e.g., doing 10bit per channel output when that's available).

mitchmindtree:
Don't do gamma conversion on `Image`s as the user should choose the suitable glium texture for their colour space.

This closes #914, cc @pedrocr

### DIFF
--- a/src/backend/glium.rs
+++ b/src/backend/glium.rs
@@ -630,7 +630,7 @@ impl Renderer {
                         },
                     }
 
-                    let color = gamma_srgb_to_linear(color.unwrap_or(color::WHITE).to_fsa());
+                    let color = color.unwrap_or(color::WHITE).to_fsa();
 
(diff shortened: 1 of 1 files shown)


pedrocr:
This didn't seem to make a difference in the image rendering and looking at the code I don't see where that Vertex color ends up being used for Image types. Is there somewhere else in the code that's doing conversions? Something seems to be downsampling f32 to u8 or similar before doing gamma but maybe that's glium itself and I'm using it wrong.

PR #927, opened 2017-01-31
Custom `Button` images for "hover" and "press" states.

mitchmindtree:
This also removes the image button's BorderedRectangle instance,
assuming that if the user is using images for buttons, they do not
require the flat rectangle to be drawn as they would have added it to
their images if it were necessary.

Image button example in idle state:

screen shot 2017-01-31 at 5 52 45 pm

and when hovered:

screen shot 2017-01-31 at 5 52 49 pm

I don't have a screenshot for press, but it's basically the same but a bit lighter.

cc @tl8roy. Is this enough to close #906? Or do you also require the custom colours for "hover"ing and "press"ing flat buttons? The reason I haven't added the custom colours behaviour with this PR is that the change could also apply to a lot of the other widgets, and I'd like to put some thought into whether there can be a "proper" solution to this that all similar widgets can also benefit from.


tl8roy:
@mitchmindtree I am using the custom colours for flat buttons at the moment, but they are on non critical screens that can wait for a "proper" solution. I can also use an image for the time being if required.

PR #930, opened 2017-01-31
Add `#![deny(unsafe_code)]` to the crate root

### LINKED ISSUE #544: Add `#![deny(unsafe)]` to lib.rs
There should really be no reason to need unsafe code in conrod due to its high level nature - adding this lint might make it clearer to contributors?

mitchmindtree:
Conrod does not require unsafe and should have no need for FFI.

This makes it clear to users that memory-safety related bugs should not
originate from conrod.

Closes #544.

### DIFF
--- a/src/backend/glium.rs
+++ b/src/backend/glium.rs
@@ -103,7 +103,11 @@ pub struct Vertex {
     pub color: [f32; 4],
 }
 
-implement_vertex!(Vertex, position, tex_coords, color, mode);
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Agh, it seems the glium backend uses a macro that directly expands to some unsafe code. I'll change this to `deny` instead of `forbid` and explicitly allow unsafe code for glium's vertex impl.

PR #931, opened 2017-02-01
Add support for OpenGL 2.1 (and in turn GLSL 120) to `glium` feature.

mitchmindtree:
@tl8roy this seems to be working nicely on my rpi2! You might have to add the following builder method to your glium Display construction:

```rust
.with_gl(glium::glutin::GlRequest::Specific(glium::glutin::Api::OpenGl, (2, 1)))
```

If I don't add that line, I get a `Failed to get root window` error.

Let me know how it goes :+1:

### DIFF
--- a/examples/all_winit_glium.rs
+++ b/examples/all_winit_glium.rs
@@ -31,7 +31,7 @@ mod feature {
             .with_vsync()
             .with_dimensions(WIN_W, WIN_H)
             .with_title("Conrod with glium!")
-            .with_gl(glium::glutin::GlRequest::Specific(glium::glutin::Api::OpenGl, (2, 1)))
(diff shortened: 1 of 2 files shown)


tl8roy:
Cool. I'll have a go first thing tomorrow.

mitchmindtree:
Btw, I still get the following error at startup on all of the examples I've tried:

```
MESA-LOADER: failed to retrieve device information
MESA-LOADER: failed to retrieve device information
```

I'm not sure what this is caused by, however it doesn't seem to affect the examples (they're running really well in terms of performance). Even `glxgears` prints this error twice so maybe it's some unrelated bug in the experimental OpenGL 2.1 driver?

tl8roy:
Good to hear about performance. This is probably the most confident I have been about getting a solution. It has been a hard slog.

Yes, that error is related to the driver, but it doesn't seem to hurt anything.

PR #935, opened 2017-02-01
Add one file hello world (glutin)

### LINKED ISSUE #932: Minimal “hello world” example
Without things like `mod support` (btw, shouldn't EventLoop be a part of conrod itself?).

suhr:
It doesn't show anything and I don't know why.

Fixes #932.

### DIFF
--- a/examples/hello_glutin.rs
+++ b/examples/hello_glutin.rs
@@ -1,0 +1,121 @@
+#[macro_use] extern crate conrod;
+
+use conrod::backend::glium::glium;
+use conrod::backend::glium::glium::{DisplayBuild, Surface};
+
+use std::time;
+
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks for the PR!

> It doesn't show anything and I don't know why.

What do you mean by this? It doesn't work?

travis is failing due to the missing `cfg` feature gates (see the other examples for how to do this).

Anyway, I'm not sure that this example is very useful. I don't think there is much benefit to including the whole `EventLoop` which already exists in the `support` module. Otherwise, everything that this example demonstrates is already demonstrated in other examples.

suhr:
> What do you mean by this? It doesn't work?

hello

No visible «Hello world!».

> Anyway, I'm not sure that this example is very useful.

It is really nice to have something you can just copypaste in your empty project, so you immediately get something working.

That's why everything is in one file, and that's why I want the example to be as simple as possible.

PR #942, opened 2017-02-10
Add support for multi-touch events. Add `Touch` and `Tap` variants to `Ui` and `Widget` events.

### LINKED ISSUE #936: Touchscreen doesn't appear to registor clicks on buttons
At the moment it appears that a touchscreen doesn't work.

This could be a thing with the touchscreen driver or the method that is used to detect clicks doesn't work for touchscreens.

FYI, the mouse moves so that the hover command works, but not the click itself.

mitchmindtree:
Also makes it so that `Button` and `Toggle` now react to `Tap`s as well as `Click`s.

This also simplifies the `Motion` input type in favour of adding a new `Touch` input type and moving it into its own variant on the relevant event types.

cc @tl8roy I think this should be working, however I'm currently out of town and don't have a touch device that I can use for testing on me. If you could test and confirm that this works that would be great!

Closes #936.

### DIFF
--- a/src/backend/piston/event.rs
+++ b/src/backend/piston/event.rs
@@ -16,22 +16,35 @@ pub fn convert<E>(event: E, win_w: Scalar, win_h: Scalar) -> Option<Input>
 
     if let Some(xy) = event.mouse_cursor_args() {
         let (x, y) = translate_coords(xy);
-        return Some(Input::Move(input::Motion::MouseCursor(x, y)));
+        return Some(Input::Motion(input::Motion::MouseCursor { x: x, y: y }));
     }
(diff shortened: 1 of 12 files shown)


tl8roy:
So the buttons work.

The X/Y widget doesn't but I don't know if that was meant to. The Frequency changer also doesn't work.
The small scroll bar makes things difficult too.

All that said, I think it is ready as an initial release. I know we only use the button side.

mitchmindtree:
Great! I only added touch event handling to `Button` and `Toggle` in this PR, but have opened #945 for tracking the rest.

PR #948, opened 2017-02-18
Sort the file navigator listings

pedrocr:
The file navigator was sorting dirs before files but not alphabetically within them. Simplify the code a bit and implement sorting of files and directories by their alphabetical order.

### DIFF
--- a/src/widget/file_navigator/directory_view.rs
+++ b/src/widget/file_navigator/directory_view.rs
@@ -119,10 +119,35 @@ fn is_file_hidden(path: &std::path::PathBuf) -> bool {
 }
 
 /// Returns true if file or directory should be displayed depending on configuration
-/// and file status (hidden or not)
-fn check_hidden(show_hidden: bool, path: &std::path::PathBuf) -> bool {
-    show_hidden | (!is_file_hidden(path))
+/// and file status (hidden or not) and extension (matching or not)
+fn check_hidden(show_hidden: bool, types: super::Types, path: &std::path::PathBuf) -> bool {
+    // Reject hidden files or directories
+    if is_file_hidden(path) && !show_hidden {
+        return false
+    }
+
+    match types {
+        super::Types::All => return true,
+        super::Types::WithExtension(valid_exts) => {
+            // We only filter files by extension
+            if path.is_dir() {
+                return true
+            }
+
+            // Check for valid extensions.
+            let ext = path.extension()
+                .and_then(|ext| ext.to_str())
+                .map(|s| std::ascii::AsciiExt::to_ascii_lowercase(s))
+                .unwrap_or_else(String::new);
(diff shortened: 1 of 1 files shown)


src/widget/file_navigator/directory_view.rs
```diff
@@ -196,48 +197,52 @@ impl<'a> Widget for DirectoryView<'a> {
...
                     entries.filter_map(|e| e.ok())
-                        .filter(|f| check_hidden(show_hidden, &f.path()))
-                        .collect()
+                        .filter_map(|f| {
+                            let path = f.path();
+                            if check_hidden(show_hidden, &path) {
+                                match types {
+                                    super::Types::All => Some(path),
+                                    super::Types::WithExtension(valid_exts) => {
+                                        // We're only after files.
```
mitchmindtree: This comment is no longer true after this PR

src/widget/file_navigator/directory_view.rs
```diff
@@ -196,48 +197,52 @@ impl<'a> Widget for DirectoryView<'a> {
...
+                                                None
+                                            }
+                                        }
+                                    },
+                                }
+                            } else {
+                                None
+                            }
+                        }).collect()
                 }
```
mitchmindtree: This section has quite a lot of rightward-drift, I quite prefer the original style that used `filter` for `check_hidden` and an early return for `path`s that are directories. I'd recommend maintaining the author's original styling in general when submitting PRs as 1. the styling generally comes down to personal preference and 2. styling changes make it harder to find the actual changes that are relevant to the PR itself. Not a big deal though, just thought I'd mention it.

src/widget/file_navigator/directory_view.rs
pedrocr: Yeah, it should be "we're only filtering files" or something like that.

src/widget/file_navigator/directory_view.rs
pedrocr: The original code did things in a slightly strange way. First it created a Vec with all the entries. Then it created an iterator that went through the Vec once selecting all the files. Then it created another iterator that went through the Vec again selecting all the files that matched. And then it joined the two iterators. In the end it was doing several Vec traversals. If I had kept that code, I would have to do a collect() on that final iterator to sort again. That's what I did original and the code was just getting too complex. If it's the rightward drift you're worried it's easy enough to just move the contents of the filter_map into it's own function together with check_hidden.

pedrocr:
On the letting the user pass the function I fully agree. This code is almost there as it already does the sort_by exactly like you describe so the only difference would be to make that sort_by closure the default and allow the user to override it.

src/widget/file_navigator/directory_view.rs
pedrocr: I've pushed a version that moves all the filtering into check_hidden, avoiding all the right drift and making the code a little simpler to read.

Issue #612, opened 2015-11-02
Provide `Ui` methods that indicate whether or not user input is currently being captured.


mitchmindtree:
The method would return `None` if the input is not captured, otherwise returning `Some(widget::Index)` with the index of the widget currently capturing the input. I.e.
```rust
if ui.is_mouse_captured().is_none() {
    // Do stuff with mouse
}

if ui.is_keyboard_captured().is_none() {
    // Do stuff with keyboard
}
```

LaylConway:
I'm currently hitting this issue as a blocker for my editor.

LaylConway:
Nevermind it seems that the following is what I needed:
```
let widget = self.ui.global_input().current.widget_under_mouse;
```

cedric-h:
```rs
    /// Which widget, if any, is currently capturing the keyboard
    pub widget_capturing_keyboard: Option<widget::Id>,
    /// Which widget, if any, is currently capturing the mouse
    pub widget_capturing_mouse: Option<widget::Id>,
    /// The widget that is currently under the mouse cursor.
    ///
    /// If the mouse is currently over multiple widgets, this index will represent the top-most,
    /// non-graphic-child widget.
    pub widget_under_mouse: Option<widget::Id>,
```
We have all of these fields on the struct ui.global_input() returns.
Can this issue be closed?

PR #952, opened 2017-02-19
[WIP] Expose widget capturing and mouse_over state

### LINKED ISSUE #611: Provide a `Ui` method that returns whether or not the mouse is over a `Widget`.
Perhaps so that it may be used like:

``` rust
if ui.widget_under_mouse().is_none() {
    // Handle other mouse stuff.
}
```

Drakulix:
Attempt to fix #611 and #612.

Currently does not compile, the reason is this line: https://github.com/Drakulix/conrod/blob/feature/input_feedback/src/widget/drop_down_list.rs#L320

I did as suggested in #611 and made `global_input` private, so this is expected.

I would like to have some feedback regarding how I should attempt to fix this:

- Leave `global_input` public and open a new issue on that subject
- Expose more of `global_input.current` via functions to satisfy the current requirements of `DropDownList` (if yes, what exactly?)
- Refactor `DropDownList` in a way this is not required anymore (if yes, how?)

### DIFF
--- a/examples/all_winit_gfx.rs
+++ b/examples/all_winit_gfx.rs
@@ -363,7 +363,7 @@ mod feature {
             }
 
             // Update widgets if any event has happened
-            if ui.global_input.events().next().is_some() {
+            if ui.global_input().events().next().is_some() {
                 let mut ui = ui.set_widgets();
                 support::gui(&mut ui, &ids, &mut app);
             }
--- a/src/tests/ui.rs
+++ b/src/tests/ui.rs
@@ -51,11 +51,11 @@ fn test_handling_basic_input_event(ui: &mut Ui, event: Input) {
 }
 
 fn assert_event_was_pushed(ui: &Ui, event: event::Event) {
-    let found = ui.global_input.events().find(|evt| **evt == event);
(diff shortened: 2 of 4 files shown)


mitchmindtree:
Thanks for the PR!

Perhaps the issue could be fixed by providing a `pub fn global_input(&self) -> &input::Global` method? That way we can still provide public access to the `global_input` field while ensuring that the user cannot mutate it and mess up behaviour. Thoughts?

Drakulix:
Of course that's also a good idea, I somehow assumed global_input in itself should be an implementation detail. But if it shall just not be modified, that is of course an easy solution.

Do you still wish to keep the `widget_under/capturing` shortcuts, if they are accessible through `global_input` anyways?

mitchmindtree:
Hmmm good question. I think it might be best to remove the new methods in favour of accessing them via the `current` field. Although the new methods do make access a little easier, providing multiple ways of accessing the same thing almost always leads to confusion heh!

Drakulix:
Done. :smiley:

Drakulix:
Tests fixed.

mitchmindtree:
Thanks for this!

PR #959, opened 2017-03-14
Improve Display implementations for errors

daboross:
This adds a case to `conrod::text::font::Error`'s `Display` implementation which calls `std::io::Error`'s `Display` implementation instead of its `std::error::Error::description()`.

This also implements `std::error::Error` and `Display` for the two error structs created for the glium backend in a very minimal way. These could be expanded to include more information if it would be relevant, but even this minimal implementation would be helpful for displaying errors more nicely than Debug!

If there's anything else needed for this, please ask!

### DIFF
--- a/src/backend/glium.rs
+++ b/src/backend/glium.rs
@@ -893,6 +893,24 @@ impl From<glium::program::ProgramChooserCreationError> for RendererCreationError
     }
 }
 
+impl std::error::Error for RendererCreationError {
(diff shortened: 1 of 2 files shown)


daboross:
Apparently I missed some tests. Yay for travis!

mitchmindtree:
This is great, thanks!

daboross:
Thanks!

Issue #839, opened 2016-10-10
#[derive(Debug)] not sprinkled all over


ksf:
Pretty much title. The lack of `Debug` impls for even transparent plain old data makes `print!`-debugging or even just poking to gain insight rather awkward, to put it mildly.

mitchmindtree:
Feel free to add the impl where you find it necessary and submit a PR :+1:

Compilation already takes a long time, I normally hold off on arbitrarily adding impls like this until someone comes across a specific need for them, as you have.

azriel91:
Turns out I need this for `Ui`, and long story short, we need the following repositories/crates to be updated in order:

* https://github.com/dylanede/stb_truetype-rs
* https://github.com/dylanede/rusttype
* this one (conrod)

I'll link the PRs as I can make them (need to update Cargo.toml for each crate that is published)

shanehandley:
I'll assume that with #961 closed via #1234 , this can be closed.

PR #961, opened 2017-03-17
Derive Debug for Ui, all other structs it depends on

azriel91:
As per title. This (partially) addresses https://github.com/PistonDevelopers/conrod/issues/839

The PR build will fail currently, as it is anticipating a release of https://github.com/dylanede/rusttype

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ path = "./src/lib.rs"
 daggy = "0.4.0"
 num = "0.1.30"
 pistoncore-input = "0.17.0"
-rusttype = "0.2.0"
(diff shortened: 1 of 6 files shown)


mitchmindtree:
Thanks for this @azriel91, I'll merge as soon as the rusttype PR merges.

daboross:
It doesn't seem like `rusttype` is being actively maintained. Would it be alright if we just did a stub `"<non-debug>"` output for fonts for now?

mitchmindtree:
@daboross yeah I think this would be fine, or perhaps we could do a manual `Debug` impl

PR #963, opened 2017-03-21
Update the widget under the cursor on widget change

daboross:
This should fix buttons not being clickable until the mouse moves when the UI is first created or when what widgets are on the screen change under the cursor.

### DIFF
--- a/src/tests/ui.rs
+++ b/src/tests/ui.rs
@@ -85,9 +85,8 @@ fn ui_should_reset_global_input_after_widget_are_set() {
     };
 
     move_mouse_to_widget(button, ui);
-    left_click_mouse(ui);
-
-    {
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Ahh one of the tests is failing and causing travis to complain. I think we should just remove that test though as the behaviour it tests for is no longer what we want. Would you mind removing it?

Btw, I just tested this with the `old_demo.rs` `DropDownList` and it now seems to be working perfectly!

daboross:
OK - I don't fully understand what the test is testing, but would something like this be better than removing it, or would it invalidate the point of the test?

mitchmindtree:
Nice, yeah I think that's fine :+1:

PR #968, opened 2017-04-06
Enable multisample anti-aliasing for glium backend

### LINKED ISSUE #967: Antialiasing of primitives
Is it possible to enable or are there plans to make it possible?

Right now, "primitives" example clearly doesn't use antialiasing, because lines that aren't parallel to 0/45/90 degrees look jagged. (especially visible on ellipses)

screen shot 2017-04-06 at 10 22 24-or8

BTW, if that matters, the GPU which I used to run the example is known to support antialiasing in OpenGL.

mitchmindtree:
Closes #967.

@sp-1234 this seems to fix the anti-aliasing in the primitives example for me. Is this what you had in mind?

### DIFF
--- a/examples/all_winit_glium.rs
+++ b/examples/all_winit_glium.rs
@@ -31,6 +31,7 @@ mod feature {
             .with_vsync()
             .with_dimensions(WIN_W, WIN_H)
             .with_title("Conrod with glium!")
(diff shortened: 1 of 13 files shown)


sp-1234:
Looks legit, I'll try to test it…

PR #972, opened 2017-04-09
Replace default hasher for HashSet and HashMap on fnv.

Object905:
Fnv hasher s more efficient for smaller hash keys such as ids, which usually used in conrod.
Since we don't need resistance against HashDoS attacks, we can use more efficient hashing algorithm.

[Comparison of hashing algorithms in rust](http://cglab.ca/~abeinges/blah/hash-rs/)
[fnv docs](https://doc.servo.org/fnv/index.html)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ path = "./src/lib.rs"
 
 [dependencies]
 daggy = "0.4.0"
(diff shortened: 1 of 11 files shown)


mitchmindtree:
Thanks!

PR #993, opened 2017-05-31
Update winit and various examples dependencies

burtonageo:
Note that this is a bit more involved than a version bump, as I had to maintain compatibility with both `glium::glutin`'s old system where each window had its own events iterator, and `winit`'s new system, where there is a global `EventsLoop`, which has an window id associated with each event.

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -102,13 +102,19 @@ impl WinitEvent for winit::WindowEvent {
             winit::WindowEvent::Focused(focused) =>
                 Some(Input::Focus(focused).into()),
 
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks a lot for this @burtonageo and sorry about the wait!

I have local branch that makes a similar set of changes, however I've been waiting on tomaka/glutin#897 getting reviewed and hopefully merged (and updating tomaka/glium#1580 accordingly) and for both to be published before going ahead with fully updating. Is there any chance you'd be interested in reviewing tomaka/glutin#897?

mitchmindtree:
Closing as we've long past these versions :+1:

PR #994, opened 2017-05-31
Update travis to build a selection of examples

burtonageo:


mitchmindtree:
Hey, thanks for the PR!

I believe the `cargo test` commands already build the examples meaning this might not be necessary, though I could be wrong! Would you mind checking this?

SuperCuber:
I think `cargo test --examples` is enough. From `cargo help test`: 
```
--examples                   Check that all examples compile
```
So instead of this I suggest adding `--examples` to all `test` lines.

mitchmindtree:
@SuperCuber hmm I think `cargo test --examples` *only* tests compiling the examples, whereas `cargo test` compiles the crate, compiles and runs all tests *and* compiles the examples. I could be wrong though!

mitchmindtree:
Going to close this as I believe the new cases are already covered by the `cargo test` commands. Thanks anyways!

PR #996, opened 2017-06-10
[WIP] SDL2 backend

jonas-schievink:
This implements an SDL2 backend that works similar to the piston_graphics backend. It uses the `sdl2` crate, the core SDL2 library, as well as SDL_gfx, a library that can draw more complex 2D shapes like polygons (core SDL is limited to rects, points and lines).

Missing:

* [ ] Event conversion is incomplete
* [ ] Scroll events get multiplied by a rather large value because conrod scrolls far too slow otherwise
* [ ] Graphical glitches in the provided example (the scrollbar is black when scrolled down, the rounded rectangle gets a thick black border when at the edge of the screen - but only sometimes)
* [ ] Handling of line caps
* [ ] DPI handling

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,11 @@ rusttype = "0.2.0"
 # - Converting piston `GenericEvent` types to `conrod::event::Raw`s.
 # - Rendering the `conrod::render::Primitives` yielded by `Ui::draw`.
 # Enables the `conrod::backend::piston` module.
+#
+# `sdl2`
+# Provides an SDL2 based primitive renderer and event handling.
+# Enabled the `conrod::backend::sdl2` module.
+sdl2 = { version = "0.30.0", optional = true, features = ["gfx"] }
 glium = { version = "0.16.0", optional = true }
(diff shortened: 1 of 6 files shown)


alanpoon:
Hi jonas, I am trying to add rust-sdl2's example "animation.rs" into your "all_sdl2.rs" example. The animation stopped, and resumed when there is input events like mouse clicks and mouse moves. Closing the window was also difficult. Here is the code. https://gist.github.com/anonymous/968661d92b330eb76df5d601b705cbb8. Do you have any solution for animation with conrod?

jonas-schievink:
@alanpoon I'm not sure what's causing your issue, but you probably shouldn't sleep in the event loop. Using `poll_iter` instead of `wait_iter` is also a bad idea since that will terminate once all events are processed. Try using a nested loop that moves your animation and other state forward, and then polls all events. Anyway, this PR isn't a good place for troubleshooting, so I suggest asking in the `#rust-gamedev` IRC channel instead.

PR #1000, opened 2017-06-17
Publish with updated daggy to fix nightly std breakage

mitchmindtree:


mitchmindtree:
Published as **0.53.0**.

PR #1012, opened 2017-07-10
Fix incorrect scissor rect in piston backend

corngood:
https://github.com/corngood/conrod/commit/c6d562325da48e5c8b6f6ad4a2e55ebf8e5cecdc is a hack for the `all_piston_window` test that shows a problem with scissoring.

Without the fix should look like this (incorrect):

image

With the fix it should look like:

image

The same test works correctly with the glium backend.

### DIFF
--- a/src/backend/piston/draw.rs
+++ b/src/backend/piston/draw.rs
@@ -246,11 +246,11 @@ fn crop_context(context: Context, rect: Rect) -> Context {
     // We start with the x and y in the center of our crop area, however we need it to be
     // at the top left of the crop area.
     let left_x = x - w as f64 / 2.0;
-    let top_y = y - h as f64 / 2.0;
+    let top_y = y + h as f64 / 2.0;
 
     // Map the position at the top left of the crop area in view_dim to our draw_dim.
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks for this @corngood and sorry about the delay! Do you remember if any of the examples were fixed by this? Just wondering if there's a quick way I can test this before merging

corngood:
@mitchmindtree I didn't find any problems with the existing examples, but you should be able to reproduce it with https://github.com/corngood/conrod/commit/c6d562325da48e5c8b6f6ad4a2e55ebf8e5cecdc.

I don't know if there's enough there to be worthy of an example.  It's probably more of a test-suite type of thing.

elrnv:
I can confirm this issue. Ran into it when setting `crop_kids()` for a canvas split (I think it has to be not the first one).

mitchmindtree:
@elrnv to clarify, does this PR fix the issue for you? If so I'll go ahead and merge

elrnv:
Yes this PR fixes the issue for me :)

PR #1015, opened 2017-07-12
Update to glium 0.17 and winit 0.7, the new winit EventsLoop and glutin GlWindow

mitchmindtree:
Here are the relevant upstream PRs

- glium/glium#1601
- tomaka/glutin#901
- tomaka/winit#126

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,8 +45,9 @@ rusttype = "0.2.0"
 # - Converting piston `GenericEvent` types to `conrod::event::Raw`s.
 # - Rendering the `conrod::render::Primitives` yielded by `Ui::draw`.
 # Enables the `conrod::backend::piston` module.
-glium = { version = "0.16.0", optional = true }
-winit = { version = "0.5.9", optional = true }
+glium = { path = "../glium", optional = true }
+#glium = { version = "0.16.0", optional = true }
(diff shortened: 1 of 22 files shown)


mitchmindtree:
Note that although this is merged, the all_winit_gfx example does not compile currently due to a discrepency in glutin versions depended on by the latest version of gfx and the latest version of glium. Once a new version of gfx is published we'll likely be able to update to fix this.

trishume:
I tried out the latest master after this got merged and unfortunately live resize still doesn't work on macOS, even though it now works in the `glutin` examples. Not sure if this is known or intended.

mitchmindtree:
@trishume thanks for pointing this out!

This is an issue with the implementation of the examples. The examples will need a re-write to get this working properly, specifically the primitives generated by a call to either `ui.draw` or `ui.draw_if_changed` need to be drawn in the callback in which `Resized` is emitted. I think it's fine to address this in a future PR. The problem mainly stems from [this issue](https://github.com/tomaka/winit/issues/219).

That said, the `all_winit_glium_threaded` example implementation should support live resizing slightly better, though at a glance it looks like it might not work perfectly smoothly - would be curious to hear how it works for you!

I'll look into how much work is involved in updating the examples to behave properly and hopefully get around to it soon!

PR #1016, opened 2017-07-13
Simplify hello_world example and support live resize

mitchmindtree:
@trishume Here's a simpler example of handling live resize that just displays some text in the middle of the window. Note however that this does not regulate the timing of the loop in any way like the other examples do.


trishume:
Awesome thanks, the code makes sense.

PR #1017, opened 2017-07-14
[WIP] Update gfx winit example

andrewcsmith:
There's one quite hacky commit I had to make to get this to compile with the current version dependency hell, but the basic goal here is to get on track with the `all_winit_gfx` example to show where I'm headed with it. I'm sure we'll be able to put most of this boilerplate in `mod backend`.

Run it with:
`cargo run --example all_winit_gfx --features "winit glutin_winit"`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,18 +48,24 @@ rusttype = "0.2.0"
 glium = { version = "0.17", optional = true }
 winit = { version = "0.7", optional = true }
 piston2d-graphics = { version = "0.21.1", optional = true }
+glutin = { version = "0.9", optional = true }
 
 [features]
 piston = ["piston2d-graphics"]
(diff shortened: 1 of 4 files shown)


mitchmindtree:
Thanks for this @andrewcsmith!

I think I'd prefer to wait for gfx to publish a new version using the latest glutin and winit, but I'll leave this open for now in case other users come across the same issue and need to use this in the meantime.

andrewcsmith:
Oh totally, the hack is just to get it to compile. I plan to cherry-pick the real example commits on a new branch eventually, but thought this would help get the ball rolling.

mitchmindtree:
Thanks again for this @andrewcsmith, looks like the gfx backend is working nicely at least since #1190 so going to close this!

Issue #1009, opened 2017-07-07
Border shows through semi-transparent widgets


andrewcsmith:
A button with a black border that itself has an alpha of 0.1 will be mostly black, even if the clear color for the whole canvas is white. It should (intuitively) show through to the background, not the border.

mitchmindtree:
Ahh true, this has been a problem for a while, I'll have a look at fixing this now

PR #1021, opened 2017-07-16
Add new `Triangles` primitive widget

mitchmindtree:
This should allow users to draw more complex UI scenes without needing
to instantiate large numbers of primitive widgets.

This now also allows users to draw gradients.

This introduces a new `Triangles` `render::Primitive` which backends
must now handle.

The glium and piston backends have been updated to handle this, however
piston's support is currently limited due to the lack of support for
unique colours per vertex.

A simple `triangles.rs` example has been added that draws two triangles
covering the window with a unique color at each point.

screenshot from 2017-07-16 21-19-03

**TODO**

- [x] Add another `render::Primitive` that describes a list of triangles with a flat colour.
- [x] Remove the `Polygon` and `Lines` render primitives in favour of the new `Triangles` render primitive.
- [x] Add `Triangle` iterators that convert the various shapes and lines into their individual `Triangle`s.

**Future Related Work** 

- [ ] Add a more flexible `triangles::Source` trait to use rather than `IntoIterator`.
- [ ] Allow for storing triangles on the stack for smaller, simple uses of `Triangles` (rather than always using `Vec`.
- [ ] Fix BorderedRectangle by using a single `Triangles` to draw it, rather than stacking two rectangles which cause transparency issues #1009.
- [ ] Rewrite `Grid`, `RoundedRectangle` `PointPath`, `Line` and `Polygon` to use `Triangles`.


mitchmindtree:
@LaylConway this removes the `Polygon` and `Lines` `Primitive` variants in favour of simpler `TrianglesSingleColor` and `TrianglesMultiColor` variants, will this be an easy transition for your custom graphics backend? Let me know if you run into any issues, hopefully this should just simplify things though!

PR #1033, opened 2017-07-25
Finish all_winit_gfx example

Eroc33:
The `all_winit_gfx` example now draws triangle primitives.
Drawing images is still left to do, but the example is now relatively usable as a base if you want to add conrod to a gfx app.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,17 +48,18 @@ rusttype = "0.2.0"
 glium = { version = "0.17", optional = true }
 winit = { version = "0.7", optional = true }
 piston2d-graphics = { version = "0.21.1", optional = true }
+gfx = { version = "0.16.1", optional = true }
+gfx_core = { version = "0.7.0", optional = true }
 
 [features]
 piston = ["piston2d-graphics"]
+gfx_rs=["gfx","gfx_core"]
 
 [dev-dependencies]
 find_folder = "0.3.0"
 image = "0.13.0"
 rand = "0.3.13"
 # glutin_gfx.rs example dependencies
-gfx = "0.16.1"
-gfx_core = "0.7.0"
 gfx_window_glutin = "0.17.0"
 glutin = "0.9.0"
 # piston_window.rs example dependencies
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,9 +57,9 @@ find_folder = "0.3.0"
 image = "0.13.0"
 rand = "0.3.13"
 # glutin_gfx.rs example dependencies
-gfx = "0.15.1"
+gfx = "0.16.1"
 gfx_core = "0.7.0"
-gfx_window_glutin = "0.15.0"
-glutin = "0.7.0"
+gfx_window_glutin = "0.17.0"
+glutin = "0.9.0"
 # piston_window.rs example dependencies
 piston_window = "0.65.0"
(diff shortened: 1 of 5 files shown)


mitchmindtree:
Just tested this locally and it seems to work nicely, nice work!

@andrewcsmith you mentioned you might be doing some work on the gfx backend - does this conflict with any of your work? Would you be happy to see this merged?

andrewcsmith:
Yeah, this is great. I've only been working on the example for now. But if some of this is stable (the drawing routines) it would make sense to me to move it to `mod backend`.

Eroc33:
I've taken a first-pass at moving some of the example code to a backend module.
I've also noticed we might be able to share some code with the glium backend, but I haven't done anything about that yet.

Eroc33:
Something I'm not sure on here is we may want to be generic over `ColorFormat` in case, but I can't see a way to define a generic pipeline. Although then we'd have to convert textures before drawing them, so it might not be worth it anyway.

mitchmindtree:
Oh thanks for going ahead and setting up a backend module!

I'll have a closer look soon, but one thing that immediately comes to mind is that we might be best to put this in a `gfx/gl` submodule, as gfx itself is not necessarily OpenGL-specific and some uses may get confused if they try to use the module with another backend like Metal for example.

Eroc33:
Indeed gfx is not gl only. I *think* however this probably should work for every gfx backend (although I haven't tested that in any way) if we swap in the correct shaders as (I think) I haven't used anything gl specific in the backend.
I'm not really sure how to do that though. Perhaps have this in `gfx/shared` and have `gfx/gl` etc. just delegate to this with the right shader.

andrewcsmith:
The [cube example](https://github.com/gfx-rs/gfx/tree/master/examples/cube) is a super simple example of code that's generic over all backends -- basically the shaders are just included in the binary in a file `source/[shader].hlsl` or whatever. Maybe there's a way to use that as a base? The draw routines are obviously the critical thing to get working here, but maybe we can roll all this out at once.

I'd suggest we just put out a cry for help on Reddit + IRC to see if anyone can write the shader, test, and screenshot the example for the various backends that we don't have available. I'm sure it would just take 15 minutes of someone's time if they know what they're doing.

Eroc33:
Unfortunately it looks like that (indirectly) uses `ShadeExt::shader_backend` which isn't released yet afaict.

andrewcsmith:
I see...so we've got this [massive gfx update coming down the line](http://gfx-rs.github.io/2017/07/24/low-level.html), plus the gfx `Application` trait that allows for easily writing generic backends, but that hasn't landed in crates.io yet. Which circle of dependency hell are we in?

Yeah, sorry, why not merge this and get the example to compile, with the knowledge that we're 💯  ok with breaking changes once the breaking changes in gfx land. Then we can make design decisions about whether to require `gfx_app`, or what, based on what the larger community seems to do.

Eroc33:
Seems like a good plan to me.

Eroc33:
Actually I realised it wasn't so easy to add images to the backend implementation as it was, but the glium backend seemed like a good way of doing it, so I refactored this and added image rendering.

mitchmindtree:
I just had a brief look over the code and ran the example again and all seems fine to me! Going to merge :+1:

PR #1037, opened 2017-08-01
Add Input::Redraw

### LINKED ISSUE #1036: Handle winit's `Refresh` `WindowEvent`
winit has a refresh event (https://docs.rs/winit/0.7.5/winit/enum.WindowEvent.html#variant.Refresh) which requires the window to be redrawn. It seems it's currently not handled by `convert_window_event` (https://github.com/PistonDevelopers/conrod/blob/fd10b27bd0aa50d4f483260d33b66657090acbc1/src/backend/winit.rs#L78-L158), probably  because the returned `Input` event doesn't have an variant for it. Would a pr for to add such an event to `Input` and handling it in `convert_window_event` be accepted?

Thomasdezeeuw:
This is a request from the backend to rerender the UI.

This also adds support to the winit backend to convert the
WindowEvent::Refresh event to new Input::Rerender event.

Closes #1036.

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -154,6 +154,10 @@ pub fn convert_window_event<W>(e: winit::WindowEvent, window: &W) -> Option<Inpu
                 Some(Input::Release(input::Button::Mouse(map_mouse(button))).into()),
         },
 
+        winit::WindowEvent::Refresh => {
+            Some(Input::Rerender),
+        },
+
         _ => None,
     }
 }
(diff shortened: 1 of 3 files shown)


Thomasdezeeuw:
@mitchmindtree Good point will send an update in a moment.

Thomasdezeeuw:
I've updated the name and also fixed the build. Unfortunately I can't test it locally because the core-graphics and cocoa crates have problem even compiling on my machine (although when I use the conrod it works fine).

PR #1048, opened 2017-08-19
Fix #1046.

### LINKED ISSUE #1046: Centered text in hello_world example lags behind window size
When resizing the window for the `hello_world` example, the position of the text lags noticeably behind where it should be for the window's current dimensions. In some cases, it takes several seconds for the text to make it to the center of the window after resizing. [Here's a video of the problem.](https://gfycat.com/responsiblefirsthandbat)

This was initially reported as part of #1042, but since this issue seems to exist even when the scaling is not broken, I've split it into its own issue.

wkunkel:
Fix #1046. Batch events together so that we don't end up lagging behind user input. Essentially reverted back to the behavior before https://github.com/PistonDevelopers/conrod/commit/e2684251ff0afd8795db2837e8ed630c0b9720fe. I didn't include the logic to wait 16ms between updates because it's my understanding that setting vsync on the window will cause the `target.finish()` call to block and limit updates to 60fps. Correct me if I'm mistaken.

### DIFF
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -45,35 +45,53 @@ mod feature {
         // The image map describing each of our widget->image mappings (in our case, none).
         let image_map = conrod::image::Map::<glium::texture::Texture2d>::new();
 
-        events_loop.run_forever(|event| {
+        let mut events = Vec::new();
 
-            // Break from the loop upon `Escape` or closed window.
-            match event.clone() {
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks for this!

This PR re-introduces the issue of running into the blocking-during-resize bug in macos (see tomaka/winit#219) but this is already a problem in all the other examples anyway, and it's probably worth not getting swamped on platforms that emit loads of events.

PR #1053, opened 2017-08-28
Updated dependencies

bvssvni:
- Bumped to 0.55.0

PR #1055, opened 2017-09-04
SDL2 backend and an example using it and gfx in all_sdl2_gfx_rs.rs

elrnv:
Also upgraded some of the dependencies to a working state.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ rusttype = "0.2.0"
 glium = { version = "0.17", optional = true }
 winit = { version = "0.7", optional = true }
 piston2d-graphics = { version = "0.21.1", optional = true }
+sdl2 = { version = "0.29", optional = true }
 gfx = { version = "0.16.1", optional = true }
 gfx_core = { version = "0.7.0", optional = true }
 
@@ -57,10 +58,11 @@ gfx_rs=["gfx","gfx_core"]
 
 [dev-dependencies]
 find_folder = "0.3.0"
-image = "0.13.0"
-rand = "0.3.13"
+image = "0.15.0"
+rand = "0.3.16"
 # glutin_gfx.rs example dependencies
 gfx_window_glutin = "0.17.0"
-glutin = "0.9.0"
+gfx_window_sdl = "0.6.0"
+glutin = "0.9.2"
 # piston_window.rs example dependencies
-piston_window = "0.65.0"
+piston_window = "0.69.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ conrod_derive = "0.1"
 daggy = "0.5.0"
 fnv = "1.0"
 num = "0.1.30"
-pistoncore-input = "0.18.0"
+pistoncore-input = "0.19.0"
 rusttype = "0.2.0"
 
 # Optional dependencies and features
@@ -65,4 +65,4 @@ gfx_window_glutin = "0.17.0"
 gfx_window_sdl = "0.6.0"
 glutin = "0.9.2"
 # piston_window.rs example dependencies
-piston_window = "0.69.0"
+piston_window = "0.70.0"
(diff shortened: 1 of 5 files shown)


mitchmindtree:
Woah thanks so much for this @elrnv and sorry for the delay!

I just cloned your fork and tried running the new example however it failed with:

```
[mindtree@mindtree conrod]$ RUST_BACKTRACE=1 cargo run --release --features "sdl2 gfx_rs" --example all_sdl2_gfx_rs
    Finished release [optimized] target(s) in 0.0 secs
     Running `target/release/examples/all_sdl2_gfx_rs`
thread 'main' panicked at 'gfx_window_sdl::init failed!: SdlError("Could not create GL context")', /checkout/src/libcore/result.rs:860:4
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:380
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:611
   5: std::panicking::begin_panic_new
             at /checkout/src/libstd/panicking.rs:553
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:521
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:497
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:92
   9: core::result::unwrap_failed
  10: all_sdl2_gfx_rs::feature::main
  11: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  12: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:458
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:59
  13: __libc_start_main
  14: _start                                                                                                   
```

Any ideas on what might be causing this? I wonder if this has something to do with my system's setup. I believe have SDL2 installed - could there be anything else I'm missing that you're aware of? I'm running an up-to-date Arch Linux using the Gnome desktop.

Btw, it looks like merging #1053 has caused some overlap in the Cargo.toml - would you mind rebasing this onto master?

elrnv:
Looks like it stopped working on my end as well after the merge. I shall investigate... BTW I noticed there is another SDL2 PR #996  floating about but I'm not sure if it works outside the SDL2_gfx context, which is why I submitted this PR in the first place.

elrnv:
There are a few things I noticed after doing some testing with the examples on macOS.
- the current gfx backend seems to have trouble with resizing on macOS. This happens on both examples: all_winit_gfx and all_sdl2_gfx_rs (from this PR).
- Running either all_winit_gfx or all_sdl2_gfx_rs on monitors with different dpi settings seems to cause problems. Namely moving the window to the other monitor causes issues with drawing.
- Other than the issues above the current example seems to work on macos now.

I also installed Arch on a VM and there are probably two potential issues:
1. In order to run sdl2 apps in wayland we need to set "SDL_VIDEODRIVER=wayland" environment variable (taken from https://wiki.archlinux.org/index.php/wayland).
2. I couldn't get it to work even with X. I always get the `gfx_window_sdl::init failed!: WindowBuildError(SdlError("Couldn\'t find matching GLX visual"))` error. Having said this I tried opening running sdl2 examples and also opening Sdl2 windows through Piston and both work fine. Perhaps it's an incompatibility issue of old gfx with the new sdl2 lib? I'm not sure.

Leaving this here for reference, but I think it would be easiest to wait for some updates from gfx.

elrnv:
Continuing in #1084

PR #1056, opened 2017-09-12
Don't push commands if there are no vertices

### LINKED ISSUE #1045: OpenGL ES: "unable to generate a triangle primitive because there are less than 3 vertices"
This was previously reported as part of issue #884. It still occurs when using OpenGL ES on Android.

I am not experienced with GL, but I was able to fix the issue by making sure that the previous command is not pushed if there are no vertices. See https://github.com/jbg/conrod/commit/cfab936b3b5b1bc6c419de5ae939f5a83b89baec

jbg:
This fixes #1045

### DIFF
--- a/src/backend/glium.rs
+++ b/src/backend/glium.rs
@@ -447,7 +447,9 @@ impl Renderer {
                 match current_state {
                     State::Plain { .. } => (),
                     State::Image { image_id, start } => {
-                        commands.push(PreparedCommand::Image(image_id, start..vertices.len()));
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks a lot for looking into this and finding a fix @jbg - much appreciated!

I think I'm going to go with #1059 as it's a little simpler but should hopefully achieve the same outcome. Please let me know if #1059 does not actually fix this and I'll reopen this PR.

PR #1064, opened 2017-09-22
Mouse Cursor Support

elrnv:
I included a usage example in the text_edit example.
The winit backend is also extended to take Conrod MouseCursor types and set the
corresponding mouse cursor using winit.

### DIFF
--- a/examples/text_edit.rs
+++ b/examples/text_edit.rs
@@ -8,11 +8,9 @@ fn main() {
 #[cfg(all(feature="winit", feature="glium"))]
 mod feature {
     extern crate find_folder;
-    extern crate winit;
     use conrod;
     use conrod::backend::glium::glium;
     use conrod::backend::glium::glium::Surface;
-    use conrod::backend::winit::WinitWindow;
     use support;
 
     widget_ids! {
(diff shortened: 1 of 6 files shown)


examples/text_edit.rs
```diff
@@ -91,6 +93,8 @@ mod feature {
             // Instnatiate all widgets in the GUI.
             set_ui(ui.set_widgets(), &ids, &mut demo_text);
 
+            display.set_mouse_cursor(ui.mouse_cursor());
```
mitchmindtree: I think I'd prefer if we made this a single plain old function in the backend module that converted a conrod `MouseCursor` to a `winit::MouseCursor` so that usage is like this:

```rust
display.set_cursor(conrod::backend::winit::convert_mouse_cursor(ui.mouse_cursor()));
```

Kind of like how the event conversion function is `conrod::backend::winit::convert_event`.

I like that this way we don't have to import the `WinitWindow` trait this way and that it's a little clearer that conrod is just converting the mouse cursors with no other magic going on behind the scenes. Also removes the ambiguity of `Display` having both `set_cursor` and `set_mouse_cursor` methods. Would be great to hear your thoughts on this!

elrnv:
100% agree, keeping the backend code clean is great :). Changes made as requested.

PR #1069, opened 2017-09-26
Added chapter 3 to tutorial

bobgates:
I've written a step-by-step very basic guide to getting a Conrod application up and running, and added the text for it to the guide. I've also put the source file into /examples/tutorial. There are a couple of screen captures in imgur - that seems to be how other people have put in images.

I did find editing markdown in a Rust file to be an interesting challenge. My editor isn't helping, because it thinks the file is Rust. Got there in the end, though.

I'd appreciate any feedback on this work so far - both English and Rust feedback! I'd like to go further and advance the tutorial by extracting boilerplate to separate files, then going into some detail on creating widgets, etc. I plan to cover some of the awkward cases that I've seen people ask for, like multiple windows, and I'm open to suggestions.

I'm also looking for suggestions for a simple app to use as an example. I'm thinking of (yet another) to-do list tool, but would consider other requests for something small but widget-heavy.

### DIFF
--- a/src/guide/chapter_3.rs
+++ b/src/guide/chapter_3.rs
@@ -1,4 +1,11 @@
 /*!
+# #[macro_use]
+# extern crate conrod;
+# extern crate find_folder;
+
+# use conrod::{self, widget, Positionable, Colorable, Widget};
+# use conrod::backend::glium::glium::{self, Surface};
+
 
 **Setting up a basic window using glium and winit**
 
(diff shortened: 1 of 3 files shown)


bobgates:
I'm not very good with GitHub. I did ask for a review of this, never heard anything and drifted off into other stuff. The files are there in my branch if you want to use them.

daboross:
Content wise it looks good (to me at least)!

There are some problems with the text blocks in doc comments. Mainly, `rustdoc` by default assumes all code blocks are rust code which is to be compiled and tested as part of the crate's test suite.

That works well for standalone examples, but causes errors like the ones in this PR when code is split up into multiple blocks.

To fix it, all the code blocks just need to be started with ` ```no_test ` rather than ` ``` `. (`no_test` is slightly better than `ignore` since ignored "tests" are still listed in test report output, while no_test aren't counted as tests at all). For excerpts from `Cargo.toml` can be marked as ` ```toml ` instead and rustdoc will also ignore those for testing.

The more intensive alternative is to create just enough boilerplate before and after each code block in hidden code to get it to compile. Lines can be hidden in rustdoc output but included in tests by prepending `#<space>` to the line. I'd not recommend this, especially when all the code is also tested in the example.

If you want to make the changes to get `cargo test` succeeding, I think this would be good afterwards. I could probably also find time to make the small changes myself and send you a patch, if that would work better.

bobgates:
Thanks daboross, I've done as you suggested, and it's passing rustdoc --test on my machine. I had a go at getting the tests to compile too, but I'm not quite on top of getting extern crate and use <package>  to work.

daboross:
Cool, it looks like travis likes it too!

For `extern crate` to work I think it just needs to be run through `cargo test` (or `cargo test --doc` to exclude non-doc tests) rather than `rustdoc --test`.

If you'd be willing to change it, I think `ignore` might be better than `txt` since it keeps the rust code highlighting, whereas `txt` doesn't do code highlighting at all.

src/guide/chapter_3.rs
```diff
@@ -0,0 +1,495 @@
...
+let ids = Ids::new(ui.widget_id_generator());
+```
+
+`widget_ids!` is, as its name suggests, a macro provided by conrod to create
+widget ids. Conrod's primary data structure contains all the widgets, and uses
+their ids to keep track of them. The `widget_ids!` macro just provides a very
+quick and easy way of defining widget names (in this case just `text`) and
+giving them ids.
+
+After defining the widget_ids, `Ids::new` creates the the widget structure.
```
Psykopear: I was following the guide and noticed a typo here, `creates the the widget structure`

mitchmindtree:
This is really great @bobgates - I'm so sorry it took me so long to get around to reviewing! I think this does a really nice thorough job of walking through the user step-by-step and also nicely explains the reasoning behind a lot of the verbosity and boilerplate involved while staying focused on getting the simple example going. There's a couple of tiny things that are a little dated but I'm happy to merge this as is and address any other issues in a follow up PR. Again, thanks so much for such a thorough write-up!

PR #1078, opened 2017-10-15
Update winit to the 0.8 series

pedrocr:


pedrocr:
I suspect this requires bumping to 0.56 as the winit API is now different and is reexported by conrod.

mitchmindtree:
This has since been addressed in #1083 so going to close this, but thanks for the contribution either way!

PR #1084, opened 2017-10-17
SDL2 backend and an example using it and gfx in all_sdl2_gfx_rs.rs

elrnv:
Continuation of #1055
This PR adds the sdl2 backend without sdl2_gfx.
To demonstrate that it works, we also add an example using gfx-rs.


elrnv:
Closing since I'm not working on this anymore. There is another attempt at this in #996 for those still interested.

Issue #1052, opened 2017-08-26
Is a vulkano backend planned?


pedrocr:
With glium development being mostly abandoned for vulkano and with hardware catching up fast it would seem the best backend going forward would be a vulkan one. Is a vulkano backend planned?

mitchmindtree:
I'll probably get around to this eventually but don't have any immediate plans. Would be more than happy to review and help if someone beats me to it!

Kurble:
I've been working on one for most of the day actually.. If it ends up working properly I can probably submit a pull request?

PR #1119, opened 2018-01-02
Vulkano backend

Kurble:
I implemented a vulkano backend for conrod, as requested in #1052.
Most code is a fairly 1:1 port of the gfx-rs backend.
It does make some assumptions about how vulkano is used, the Renderer::fill and Renderer::draw methods take an AutoCommandBufferBuilder and return a new one.

### DIFF
--- a/src/backend/vulkano.rs
+++ b/src/backend/vulkano.rs
@@ -550,7 +550,7 @@ impl Renderer {
                     //
                     // Texture coordinates range:
                     // - left to right: 0.0 to 1.0
-                    // - bottom to top: 1.0 to 0.0
+                    // - bottom to top: 0.0 to 0.1
                     // Note bottom and top are flipped in comparison to glium so that we don't need 
                     //  to flip images when loading
                     let (uv_l, uv_r, uv_t, uv_b) = match source_rect {
@@ -558,8 +558,8 @@ impl Renderer {
                             let (l, r, b, t) = src_rect.l_r_b_t();
                             ((l / image_w) as f32,
                              (r / image_w) as f32,
-                             (b / image_h) as f32,
-                             (t / image_h) as f32)
+                             (t / image_h) as f32,
+                             (b / image_h) as f32)
                         },
                         None => (0.0, 1.0, 0.0, 1.0),
                     };
@@ -569,7 +569,7 @@ impl Renderer {
                         let x = (x * dpi_factor / half_win_w) as f32;
                         let y = (y * dpi_factor / half_win_h) as f32;
                         Vertex {
-                            pos: [x, 1.0-y],
+                            pos: [x, y],
                             uv: t,
                             color: color,
                             mode: MODE_IMAGE,
@@ -578,7 +578,7 @@ impl Renderer {
 
                     let mut push_v = |x, y, t| vertices.push(v(x, y, t));
 
-                    let (l, r, b, t) = rect.l_r_b_t();
+                    let (l, r, t, b) = rect.l_r_b_t();
 
                     // Bottom left triangle.
                     push_v(l, t, [uv_l, uv_t]);
--- a/src/backend/vulkano.rs
+++ b/src/backend/vulkano.rs
@@ -1,0 +1,760 @@
+use std;
+use std::sync::Arc;
+
+use text::{rt,GlyphCache};
+use render;
+use color;
+use image;
+use {Rect, Scalar};
+
+use vulkano::command_buffer::AutoCommandBufferBuilder;
+use vulkano::pipeline::viewport::Scissor;
+use vulkano::instance::QueueFamily;
+use vulkano::device::*;
+use vulkano::pipeline::*;
+use vulkano::descriptor::descriptor_set::FixedSizeDescriptorSetsPool;
+use vulkano::buffer::cpu_pool::CpuBufferPool;
+use vulkano::buffer::BufferUsage;
+use vulkano::buffer::CpuAccessibleBuffer;
+use vulkano::framebuffer::*;
+use vulkano::image::*;
+use vulkano::format::*;
+use vulkano::sampler::*;
+use vulkano::pipeline::viewport::Viewport;
+use vulkano::command_buffer::DynamicState;
+
+/// A `Command` describing a step in the drawing process.
+#[derive(Clone, Debug)]
+pub enum Command<'a> {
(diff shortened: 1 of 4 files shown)


mitchmindtree:
Wow, this is amazing stuff. Thanks so much for submitting this @Kurble and sorry for the delayed response! I've been a bit swamped with moving and commercial projects atm and haven't had as much time as I'd like for my OSS stuff. Hoping to dive back in properly within the coming couple weeks.

If you or anyone else would like to make an `all_winit_vulkano.rs` example (that mimics the `all_winit_glium.rs` and `all_winit_gfx.rs` examples) that would make it much easier for me to go ahead with testing and merging :+1:

src/backend/vulkano.rs
```diff
@@ -550,16 +550,16 @@ impl Renderer {
                     //
                     // Texture coordinates range:
                     // - left to right: 0.0 to 1.0
-                    // - bottom to top: 1.0 to 0.0
+                    // - bottom to top: 0.0 to 0.1
```
DancingBard: @Kurble 0.0 to 1.0, or?

Kurble: Totally, nice find :)
I still need to take some time to write out the example anyway

OtaK:
Hey! I'm quite interested in this PR, what's missing to merge it in? I'd guess unit tests for vulkano backend, right?

Cheers for the amazing work @Kurble

Kurble:
It's still missing an all_winit_vulkano.rs example, which I haven't felt motivated to do yet (due to other unrelated stuff I'm working on). If anyone is interested in making one I can post the vulkano plumbing I used to get it up and running in my own project.

DancingBard:
@Kurble Yes, please post the plumbing!

DancingBard:
@ex-rzr Hey, sorry for the late reply. Sadly it's negative: I've not found time to tinker with this part of the app any further and I am in the same boat as you. I can render the rectangles, but if I leave `update_texture` as it is, I get:

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AutoCommandBufferBuilderContextError(ForbiddenInsideRenderPass)', libcore/result.rs:945:5

coming from within `update_texture`.

I had a very short look at https://docs.rs/vulkano/0.9.0/vulkano/command_buffer/struct.AutoCommandBufferBuilder.html#method.copy_buffer_to_image_dimensions then, but couldn't find anything overly suspicious. Will check again, once I have more time and else fallback to asking @tomaka for some insight here.

Trying to get an image to show I also fail, but that may just be me:

> AccessError { error: BufferNotInitialized, command_name: "vkCmdBindDescriptorSets", command_param: "Image bound to descriptor 0 of set 0", command_offset: 7 }', libcore/result.rs:945:5

Again, this could very well be me, I am successfully drawing textures elsewhere, and maybe I am not setting up something correctly to work with this conrod-vulkan-glue :)

I also want to go back on an earlier statement, with `update_texture` stripped down to doing nothing, the `title_bar` works as well on a canvas.

DancingBard:
@ex-rzr Just in case you're still interested.

I mentioned a different error message and for the text rendering the issue was on my side and there alone!

My problem was simply that I was calling `begin_render_pass` before `fill` of the "`GuiRenderer`". So, yes, text is working just fine right now.

Now I am working on getting the rust logo to render correctly, as there I still have that issue I mentioned above. (But only if it comes into view!)

I hope that I'll solve that today and then I have a working `all_winit_vulkano.rs` and the nice vulkano-devs could, if to their liking, maybe, merge :)

DancingBard:
So, I got everything working for me locally, but only with `winit 0.15` which I can't push, because it breaks `glium` (which relies on old version of `glutin`, which in turn uses old winit)

With older winit, I can't get it to run, or rather I don't want to put more time into it, as I personally try to stay up to date with `winit`.

Maybe it's time to split up the backends into different crates? See my proposal: https://github.com/PistonDevelopers/conrod/issues/1183

ex-rzr:
@DancingBard, thank you for the example!

It works for me, however the Rust logo can be rendered correctly only after these changes:

```patch
diff --git a/src/backend/vulkano.rs b/src/backend/vulkano.rs
index 5bad8c0..94e8de0 100644
--- a/src/backend/vulkano.rs
+++ b/src/backend/vulkano.rs
@@ -561,15 +561,12 @@ impl Renderer {
                              (t / image_h) as f32,
                              (b / image_h) as f32)
                         },
-                        None => (0.0, 1.0, 0.0, 1.0),
+                        None => (0.0, 1.0, 1.0, 0.0),
                     };
 
                     let v = |x, y, t| {
-                        // Convert from conrod Scalar range to GL range -1.0 to 1.0.
-                        let x = (x * dpi_factor / half_win_w) as f32;
-                        let y = (y * dpi_factor / half_win_h) as f32;
                         Vertex {
-                            pos: [x, y],
+                            pos: [vx(x), vy(y)],
                             uv: t,
                             color: color,
                             mode: MODE_IMAGE,
```

Perhaps, the old version of winit is the cause of my issues too. I'll check my code with the latest winit, if it won't help I'll compare/copy&paste my code for initialization, event loop with yours.

ex-rzr:
Btw, `vulkano.rs` has CRLF line endings (while other conrod files have LF). I think this should be fixed before merging.

ex-rzr:
Ok, I know in what conditions my code crashes: I create all widgets before `ui.draw()`
```rust
loop {
    ...
    {
        let ui = &mut ui.set_widgets();
        ...
    }
    ...
    let primitives = ui.draw();
    let command_buffer_builder = renderer.fill(command_buffer_builder, &image_map, viewport, primitives);
    ...
    events_loop.poll_events(|event| {
    ...
}
```

But if I create widgets at the end of the loop iteration (as it's done in all_winit_vulkano and all_winit_gfx) then it works great:
```rust
loop {
    ...
    let primitives = ui.draw();
    let command_buffer_builder = renderer.fill(command_buffer_builder, &image_map, viewport, primitives);
    ...
    events_loop.poll_events(|event| {
    ...
    {
        let ui = &mut ui.set_widgets();
        ...
    }
}
```

I'm not sure yet why this is happening, perhaps this is related to the order of swapchain recreation, previous_frame_end.cleanup_finished()...

mitchmindtree:
Closed by #1225.

PR #1123, opened 2018-01-22
bump glium / winit to fix #1120

FauxFaux:
Fixes #1120.

PR #1110, opened 2017-12-15
Fixed missing lifetime marker on GFX backend

OtaK:
Details in #1109

PR #1124, opened 2018-01-22
add lifetime to GlyphCache / Renderer

FauxFaux:
While trying to get `glium` fixed in, it was [noted](https://github.com/PistonDevelopers/conrod/pull/1123#issuecomment-359444069) that the gfx backend was broken. The gfx backend is /additionally/ broken by changes to `rusttype`, which have apparently been missed by CI.

This is the minimal change to get the code to compile. It doesn't seem harmful. I didn't think it through at all.

### DIFF
--- a/src/backend/gfx.rs
+++ b/src/backend/gfx.rs
@@ -146,9 +146,9 @@ impl Vertex {
     }
 }
 
-pub struct Renderer<R: Resources>{
+pub struct Renderer<'font, R: Resources>{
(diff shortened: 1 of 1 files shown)


daboross:
https://github.com/PistonDevelopers/conrod/pull/1110 and this seem to do different things to accomplish the same goals: might be good to compare? I have no idea.

FauxFaux:
An exact duplicate of #1110, with different spelling. My bad.

PR #1127, opened 2018-01-25
Update the remaining dependencies

daboross:
This updates piston2d-graphics, image, rand, and piston-window.

Everything that worked before for me works now. In order to run things, I did have to change `multisampling` to 0 in all examples, and change the gfx color format from `Srgba8` to `Rgba8`, but I had to do those things on this computer before this update as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ rusttype = "0.3.0"
 # Enables the `conrod::backend::piston` module.
 glium = { version = "0.20", optional = true }
 winit = { version = "0.10", optional = true }
-piston2d-graphics = { version = "0.23.0", optional = true }
+piston2d-graphics = { version = "0.24", optional = true }
 gfx = { version = "0.17", optional = true }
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Agh true, is there a glium issue open for this multisampling issue?

Thanks a lot for this!

daboross:
I'll open one - pretty sure it's just an issue with my local configuration though, it was a change in my system which caused it, not any update.

daboross:
Err, huh. I could have sworn they didn't work without changing the multisampling setting, but today everything is functional. In either case I'm pretty sure it was something messed up about my system...

Either the latest updates I've installed have fixed it, or it wasn't actually a problem and I was imagining it.

daboross:
@mitchmindtree I think I've figured out the multisampling issue: it occurs only when I'm using wayland with the winit backend set to X11. I think it's more of a problem with my setup than anything else, though, since no glium applications run well in wayland in either mode (x11 compat or directly with wayland).

PR #1144, opened 2018-03-07
Update Syn to 0.4 for conrod_derive

alanpoon:
The reason update syn dependencies is because of this reason. dtolnay/syn#364


mitchmindtree:
Thanks a lot for this @alanpoon and sorry about the delay! Really appreciate you taking the time to submit this.

I've just tested this locally and it looks good to go! Seeing as this shouldn't cause any downstream breakage I'll give this a tiny version bump and publish it now :+1:

mitchmindtree:
`conrod_derive` **0.1.1** has been published.

PR #1151, opened 2018-03-26
Update `conrod_derive` to `0.1.1` for recent `syn` update.

mitchmindtree:
This should not break any downstream code as usage of the derive macros
themselves should not change at all. That said, if you experience any
issues due to this please let us know so I can correct it!


mitchmindtree:
`conrod_derive` has been published as **0.1.1**.

PR #1160, opened 2018-04-08
Added tap event for list_select

alanpoon:
Currently on mobile, conrod's list_select widget cannot react to tab's event.


mitchmindtree:
Thanks a lot @alanpoon, but before merging I just want to clarify that you meant to say "tap" rather than "tab" - is that correct? If so, could you rename the "tab" variable in the added code to "tap"? Thanks!

mitchmindtree:
Thanks!

PR #1181, opened 2018-06-03
Bump `syn` to 0.14, `quote` to 0.6

hcpl:
Also adds `proc-macro2` dependency because `quote::Tokens` has been replaced by `proc-macro2::TokenStream`. The `^0.4.2` version requirement is because this library uses the `Ident` == `AsRef<str>` implementation.


mitchmindtree:
Thanks!

PR #1195, opened 2018-08-17
Don't crash in case there's an image missing

pedrocr:
In multi-threaded programs it's common to have one thread changing state and another displaying. It's easy in that situation to have an image already be missing from the map while a frame that still depends on it is still being rendered. Fix that by just not displaying that image. Might as well have a visual glitch instead of a crash.

### DIFF
--- a/src/backend/glium.rs
+++ b/src/backend/glium.rs
@@ -751,50 +751,52 @@ impl Renderer {
 
                     let color = color.unwrap_or(color::WHITE).to_fsa();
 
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks @pedrocr and sorry for the delay, this sounds reasonable to me!

pedrocr:
I fixed the actual issue and avoided the visual glitch thanks to that. I guess hard crashing made me more motivated to fix things :) But it's nice to not have this pitfall for a minor issue.

Issue #1183, opened 2018-06-24
[SUGGESTION] Split up backends into own crates and leave a conrod_core to allow fine-grained dependency-management


DancingBard:
**What issue I want to solve:**

I was working to set up an example for the vulkano backend by @kurble from https://github.com/PistonDevelopers/conrod/pull/1119 as requested by @mitchmindtree in https://github.com/PistonDevelopers/conrod/pull/1119#issuecomment-360098502.

When I ran the existing examples, I figured that it would be nice to also use an up to date version of `winit`, as I use it in my personal project. Winit is shaping up nicely and conrod uses 0.9, while 0.15 is the current release.

That is because (correct me if I am wrong here!) `glium` uses `glutin` which in turn uses an old version of `winit`.

Updating to `winit 0.15` worked fine for my `all_winit_vulkano.rs`-example, but it broke the `glium` ones, because the newest version of `glium` (at the time of writing this: https://crates.io/crates/glium/0.21.0) relies on `glutin 0.14` https://crates.io/crates/glutin/0.14.0, which in turn only uses `winit 0.12`

On the other hand, going back to winit `0.12` my `all_winit_vulkano.rs` doesn't work, because 

> `std::sync::mpsc::Sender<(winit::os::unix::x11::WindowId, i16, i16)>` cannot be shared between threads safely

Personally I don't intend to put more effort into supporting an outdated version of `winit`, just to get the `all_winit_vulkano.rs` example running because of some unrelated `glium` "issue".

**The solution I want to propose:**

Split the backends (WITH the[ir] `examples`) into different sub crates which can then specify their own dependencies.

That way we could use the newer `winit` version in the vulkano-backend (examples) and still have working `glium` backend that relies on an older version at the moment.

mitchmindtree:
Breaking conrod into a more modular set of crates is something that has been on my mind for a while #754. Perhaps this can be done as a part of #1103. I've been too busy to make the move recently, but will get some time in the coming weeks as we're travelling to south-east asia for some serious R&D soon.

As for the `glium`/`winit` conflict issue, I feel you - I've opened a PR to update glium here glium/glium#1693 - if you (or anyone reading this) could perhaps leave a review/comment confirming that the PR works fine, this might make it easier for tomaka to just merge it.

ianrrees:
[...]

Issue #754, opened 2016-07-29
Stripping Conrod down to a stable core


mitchmindtree:
Over the past few months I've been attempting to strip conrod down to a solid core. Many unnecessary dependencies have been removed and backend-specific code has been slowly untangled. However, there are still quite a few steps remaining and decisions to be made.

The two most unstable parts of conrod at this point are:

- the provided non-primitive widgets. These are often refactored or simplified as other widgets are added or API improvements are made.
- the backend feature-gated modules. Each of these are simply a collection of helper functions that simplify the process of getting up-and-running with conrod for different frequently requested backends. Each of these backends are unstable and prone to API changes that will no doubt require updating these modules time to time.

Apart from these two areas, conrod is getting close to the point where our primary focus will be stabilising and simplifying the existing functionality in order to work towards reaching a 1.0.0-alpha release.

I've been considering the idea of refactoring conrod into the following crates:

- **conrod** - contains only the core framework (ui, widget graph, theme, widget trait, primitives, etc).
- **conrod-widgets** - contains all non-primitive widgets.
- **conrod-piston**, **conrod-piston_window**, **conrod-glutin**, **conrod-glium** - each of the currently feature-gated modules moved into their own crates

This way we might have a chance of stabilising conrod itself in the near future, while continuing to build new widgets and improve/adapt support for various backends.

There's still a lot to be done (see the [1.0.0 milestone](https://github.com/PistonDevelopers/conrod/milestone/2)) but just thought I'd share my thoughts on this in the meantime.

bugabinga:
hi

i think the split into conrod/conrod-widgets is pretty useful for a lot of reasons. 

- consumers of conrod can choose to opt out of (presumably) large crate
- the separation most likely will keep the api boundaries between primitives and widgets clearer, making sure custom widget development is supported well
- a separate crate for widgets now paves the way to split them up even further down the line. should that make sense.

however, i am unsure what __actual__ stability the split of the backends will provide for conrod application devs.
any real application will have to use some backend. when updating conrod, an update of the backend is mandatory. from the perspective of _stability_, this seems to be the same as now. it is unfortunate, but simply a reflection of the state of these backends right now.
unless you meant, that conrod-core will supposed to be backwards compatible with some number of backend versions back. however, i think this would be too large a burden on conrod devs.

this is not to say that separate crates for the backends (or one separate for all) is not a good idea. only, when you say _stability_, as a conrod app dev i think: "how likely is something to break when updating?".

[...]

PR #1203, opened 2018-09-19
Attempt to split conrod backend from the core (see issues #754, #1183)

DancingBard:
@mitchmindtree (et al) this is an attempt to split conrod into individual backends as per https://github.com/PistonDevelopers/conrod/issues/1183


mitchmindtree:
Epic work @DancingBard! It looks like I'll be needing conrod in a project within the coming weeks so just wanted to let you know I'll get around to taking a look at this and #1204 soon - it's just a bit big for me to look at on a Sunday night before bed :)

DancingBard:
@mitchmindtree No problem.

I wanted to apologize though: I am no good with travis-ci it seems. (First project I ever tried to configure this on - and failed miserably)

Also note that we could (and maybe should) go one step further, as mentioned by your original issue, and split conrod_core and conrod_widgets

PR #1204, opened 2018-09-19
Conrod vulkano

Abendstolz:
Builds upon #1203 and tries to resolve issues with #1119 (namely introduces the missing example)


DancingBard:
@Kurble I did not add you to the cargo.toml authors, because you didn't do it in your original pr. But maybe now is the time to step forward and take the glory that is rightfully yours? ;)

mitchmindtree:
Hey @Abendstolz! Thanks again for this, I'm just working on a branch which updates this for #1223.

I noticed that you didn't use  the `vulkano_win` crate and instead have your own whole `vulkano_winit_glue` wrapper module going on in the examples `support/` module! Was this because `vulkano_win` didn't have what you were after at the time? Or did it not exist at that point maybe? Or maybe you missed it haha? Just want to double check with you and get your thoughts before plowing ahead and trying to replace it with using the `vulkano_win` crate.

Abendstolz:
@mitchmindtree exactly. If I remember correctly - and I certainly don't remember the exact versions - winit had changed (part of) their (event) API and vulkano-winit wasn't there yet.

mitchmindtree:
Ahh yep thanks for that, I've managed to switch out the glue modules in favour of `vulkano_win` and `conrod_winit`, so now the only extra "support" stuff is the window wrapper that simplifies resizing etc :+1:

Abendstolz:
@mitchmindtree and that resizing thing can and should definitely be improved. I didn't get around to it yet.

At least on my 4k desktop I have some weird problem anyway, where I have to resize the window each time manually once, in order for conrod to get the appropriate dpi-factor.

But first things first :)

mitchmindtree:
> where I have to resize the window each time manually once

Ahhh interesting, I wonder if that's to do with an old version of winit rather than your implementation? At a glance what you've got here looks OK. I think I remember seeing an issue about that there a little while back. Anyway, we'll find out soon enough.

mitchmindtree:
Closed via #1225.

PR #1214, opened 2018-11-16
Update dependencies

bvssvni:
- Bumped to 0.62.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod"
-version = "0.61.1"
+version = "0.62.0"
 authors = [
     "Mitchell Nordine <mitchell.nordine@gmail.com>",
     "Sven Nilsen <bvssvni@gmail.com>"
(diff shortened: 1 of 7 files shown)


bvssvni:
I think there is a dependency conflict between two different versions of Glutin, but I managed somehow to solve it. Please check it carefully before merging.

bvssvni:
I tried building it with different Cargo features to see if anything broke. Have not gotten any breaks so far.

atouchet:
Shouldn’t winit be updated as well?

bvssvni:
@atouchet This results in a dependency conflict. I think it's because Conrod targets all these backends with dependency versions that need to match each other.

mitchmindtree:
@bvssvni thanks for this! Sorry about the delay!

Like @0e4ef622 mentioned we probably can't merge this as it means the `all_winit_gfx` example would now depend on the `glium` feature being enabled. Do you happen to remember what other crate the glutin version conflict was arising from? No worries if not - I'll be looking into making a new conrod release with vulkano support soon (we need it in a downstream project) so I'll be sure to refer to this, investigate the glutin conflict and bump the piston backend at the same time.

@atouchet yes like bvssvni mentioned this is because the latest released version of glium is currently on an old version of winit. I think a nice easy fix would be for us to separate the dependency between the winit and glium features so that user's using winit with non-glium backends don't get slowed down by glium's slow update process.

PR #1220, opened 2018-12-04
Update rusttype.

dvc94ch:
updates rusttype to 0.7.2


mitchmindtree:
Thanks!

This is already included in #1214, will get around to this very soon, see my comment [here](https://github.com/PistonDevelopers/conrod/pull/1214#issuecomment-443134267).

PR #1241, opened 2018-12-26
WIP widget split + 2018 edition

Abendstolz:
This Pull request is WIP!

I've:
1. Split out _most_ non primitive widgets from core to it's own crate, named `conrod_widgets`
2. Migrated `conrod_derive`, `conrod_core` and (new) `conrod_widgets` to rust 2018 edition

Tested with `all_winit_vulkano` in `backends/vulkan`

**Todo:**

- Talk with @mitchmindtree to clarify whether rust 2018 edition is even okay for `conrod`
- Test and fix other backends
- Think about splitting out remaining non-primitive widgets from `conrod_core` (`bordered_rectangle`, `button`, `rounded_rectangle`)


backends/conrod_example_shared/src/lib.rs
```diff
@@ -4,13 +4,14 @@
 //!
 //! - `pub struct DemoApp` as a demonstration of some state we want to change.
 //! - `pub fn gui` as a demonstration of all widgets, some of which mutate our `DemoApp`.
-//! - `pub struct Ids` - a set of all `widget::Id`s used in the `gui` fn.
+//! - `pub struct Ids` - a set of all `conrod_widgets::Id`s used in the `gui` fn.
```
Abendstolz: @Abendstolz fix me: a replace mistake

jkelleyrtp:
Is there still a desire to separate the widgets from conrod core? I'd love to see this implemented to get better templates on how to create custom widgets.

mitchmindtree:
Yes I'd still like to see this happen, but won't get the chance to do this myself anytime soon. This PR looks like it got fairly close though!

Re 2018 - maybe it's worth opening a new issue about this for visibility to other users. I personally don't mind too much, and think it would be nice to update to 2018, but it would be nice to see that happen in a separate PR either before or after this one.

Also another thing that crosses my mind, it might be nice to `use conrod_widgets as widgets;` in the examples?

Abendstolz:
As I opened this PR half a year ago let me say that I am still willing to put work into it, but only if I am sure this gets enough attention and merged (after incorporating any feedback of course) in time. Maybe conrod needs more maintainers?

I agree that a split of the work regarding 2018 edition and splitting the widget out makes sense, I did that while I was diving through the codebase and it probably resulted in too many changes for one PR.

PR #1248, opened 2019-02-07
[conrod_vulkano] Replace incorrect gfx comment in vulkano example

singron:
The old comment made it sound like conrod is somehow simultaneously using OpenGL and Vulkan.


mitchmindtree:
Thank you!

PR #1249, opened 2019-02-07
Remove outdated comment about BorderedRectangle

singron:
The comment referes to the bug in #1009, which was fixed in #1025.


mitchmindtree:
Thanks for the PRs @singron!

PR #1252, opened 2019-02-13
Fix color space and GPU sync issues in `all_winit_vulkano.rs` example

mitchmindtree:
This fixes an issue where the example would not always pick an sRGB
format and color space, even though this is what conrod requires.

Also fixes a bug on Mac where the CPU would out-pace the GPU by ensuring
that we wait for the previous frame presentation to complete before
presenting the next one.


mitchmindtree:
cc @freesig this should fix the `all_winit_vulkano.rs` example on your mac - would you mind fetching this branch and checking for me?

PR #1253, opened 2019-02-13
Fix scissor in `conrod_vulkano` backend.

mitchmindtree:
Fixes an issue with the scissor rect calculation in the `conrod_vulkano`
backend caused by an inversion on the y axis (possibly left over from
the port from the glium backend).

Also adds a `list_select.rs` example to the vulkano backend that was
used to check the correctness of the scissor. It is an exact port of the
glium backend `list_select.rs` example.


mitchmindtree:
cc @freesig can you confirm this branch of conrod fixes your list issues downstream? Might have to use `[replace]` to override the conrod version in nannou.

freesig:
This is working on my machine

PR #1255, opened 2019-02-16
Make Position::Relative(Relative::Place, None) work with an explicit parent

Imberflur:
When creating a new widget with a `Position::Relative(Relative::Place, None)` the widget should be placed on its parent. However, if the parent is specified with `Widget::parent()` this does not work because within `Widget::set_widget()` the position is calculated before `ui.maybe_current_parent_id` is updated. Instead, the previous widget's parent ends up being used to calculate the position. For example, this won't work:
```rust
Image::new(image_id)
    .w_h(w, h)
    .parent(parent_id)
    .mid_bottom()
    .set(id, ui_cell);
```
This changes `calc_xy()` to take the `maybe_parent_id` of the widget as an input and use it to properly determine the parent of the widget.

An alternative fix would be to add a function that allows updating `ui.maybe_current_parent_id`.

### DIFF
--- a/conrod_core/src/ui.rs
+++ b/conrod_core/src/ui.rs
@@ -940,6 +940,7 @@ impl Ui {
     /// should target a **Widget**'s `kid_area`, or simply the **Widget**'s total area.
     pub fn calc_xy(&self,
                    maybe_id: Option<widget::Id>,
+                   maybe_parent_id: Option<widget::Id>,
                    x_position: Position,
                    y_position: Position,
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Thanks so much for this @Imberflur, I imagine this would have been tricky to track down! I will aim to test this locally and that all the examples still behave as expected before merging soon.

mitchmindtree:
Tested locally and everything seems to be behaving as expected, thanks again!

PR #1260, opened 2019-03-10
Changing the way dragging is handled

csirkeee:
Hi!

When trying to use Conrod for an experimental project, I started out with the "canvas" example. I noticed that dragging the floating canvases was a bit wrong. If I held one for a long time, moving the mouse all around quickly, the mouse could sometimes slide relative to the floating box, and it would let go of it if it slid completely off. (The "graph" example has a similar problem, but because that doesn't check a `drag_area`, you can keep holding a node even when the mouse cursor moves far away from it.)

I'm not sure where the difference comes in, but adding together the many `delta`s is less reliable than just taking the difference between the `origin` and `to` of the `drag`. But to use that I need to know where the widget started from, so I needed to add a new field to CommonState to help with that.

This version worked well for me, even when I slowed the test down to 1 FPS it did what I expected.

### DIFF
--- a/conrod_core/src/graph/mod.rs
+++ b/conrod_core/src/graph/mod.rs
@@ -4,7 +4,7 @@
 //! The primary type of interest in this module is the [**Graph**](./struct.Graph) type.
 
 use daggy;
-use position::{Axis, Depth, Rect};
+use position::{Axis, Depth, Point, Rect};
 use std;
 use std::any::Any;
 use std::ops::{Index, IndexMut};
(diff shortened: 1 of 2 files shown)


csirkeee:
Added one more change: Widgets should ignore dragging if they are graphics for another widget. I think this is universal, but if you don't like it, revert this part.

mitchmindtree:
Thanks so much @csirkeee! I'll clone this and test it locally next time I get a chance :+1:

mitchmindtree:
I just tested this locally and everything seems to behave as expected. Would you mind rebasing and resolving the conflict in widget/mod.rs? Sorry about that, I think you just need to update the call to `ui.calc_xy` to include the parent id argument as introduced in #1255 which I just merged.

Once that's done I'll be happy to merge!

csirkeee:
I resolved the conflict 😃 

One more question: in my project, I want a widget that is bound within its parent when dragged around. This is possible to do in user code, outside the library, but if you like, I could also add it as default behaviour in the library? So if a draggable widget has a parent it's always bound inside it. I'll make a separate pull request for that if you think it's a good idea.

PR #1261, opened 2019-03-18
Update local helper macro use

csirkeee:
Change local helper macro use according to
https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html

This way, if someone imports in just `builder_methods`, and not `builder_method`, their code will still compile


mitchmindtree:
Looks good to me! Thanks again :)

PR #1262, opened 2019-03-30
Support different versions of winit for different backends

mitchmindtree:
One ongoing struggle with conrod has been ensuring that all backend
winit dependencies support the same version of winit so that the
`conrod_winit` crate functions could be shared between each of the
different backends. Previously, only `glium` and `gfx` backends had to
be kept in sync. With the recent addition of the `vulkano` backend the
problem of keeping all backend winit dependencies up-to-date is
starting to get out of hand.

This change moves the `conrod_winit` conversion functions behind public
macros, allowing for the user to generate the conversion functions
necessary for whatever version of winit they happen to currently have.

This is a **breaking change** due to the removal of the `conrod_winit`'s
conversion functions and the removal of its `winit` and `conrod_core`
crate dependencies. See the updated `conrod_vulkano`, `conrod_glium` and
`conrod_gfx` backends and their examples to how to update to this change
(required changes should be minor).


mitchmindtree:
Travis seems to be failing on a rustc bug on the latest nightly branch - see rust-lang/rust#59502. It looks like there's already a PR up to fix the issue. Otherwise, stable and beta seem to be working fine.

Issue #1118, opened 2018-01-02
Any plans in adding support on Webassembly?


alanpoon:
I am trying to pass the Vertex{pos,uv,color,mode} generated by conrod to javascript webgl2 using wasm. I am using the conrod::backend::glium's vertex and fragment sources. In webgl2, I am using gl.vertexAttribIPointer to add an array buffer of Uint32Array and specify the type to gl.UNSIGN_INT. In glsl I am trying write to "in uint mode;", assigning an array of multiple 1 (s), but it still matches with uint(0) in the glsl. Does anybody know how to resolve this? 

Conrod --feature gfx_rs does compile in wasm32-unknown-unknown with the need of rand 0.3.18 and with a patch update. Not with glium. I think it is time for glium to start to support wasm since it claims to be able to support webgl. If not, I have to do some manual work myself.

mitchmindtree:
Hey @alanpoon, thanks a lot for the issue! If the issue is currently within the glium crate itself, perhaps this issue might be better of under the glium repository? I'm happy to keep this open though to discuss the state of using conrod with WASM :+1:

lf94:
@alanpoon did you end up creating the issue there? Could you post the link here?

alanpoon:
Nope, I was thinking of writing a new backend for webgl. Stuck at this issue: https://github.com/oussama/webgl-rs/issues/2. Glium is very under developed in webgl.

alanpoon:
there is a gui library that supports wasm. https://github.com/ryanisaacg/quicksilver
i think we can learn from it and port into conrod.

PR #1264, opened 2019-04-09
Fix a panic when using conrod targetting WASM

sebcrozet:
Hi!

I'm in the process of integrating `conrod_core` to my graphics engine [kiss3d](https://crates.io/crates/kiss3d). You can see the current result of this unfinished integration (after patching `conrod_core` with this PR) there: http://www.crozet.re/kiss3d_ui/

Currently, `conrod_core` is unusable when targeting WASM because it will cause a panic whenever the user presses a button. This is due to the call to `std::time::Instant::now()` which will panic on WASM because it requires a syscall. To fix this, I've created the [instant](https://crates.io/crates/instant) crate that provides the type `instant::Instant` which is:

- A type emulating the behavior of `std::time::Instant` using the javascript function `performance.now()` when compiling for WASM **and** when its feature `stdweb` or its feature `wasm-bindgen` is activates.
- A type alias for `std::time::Instant` otherwise.

I don't know if this is wanted here, but this PR replaces all occurrences of `std::time::Instant` by `instant::Instant` and adds the two features `stdweb` and `wasm-bindgen` to enable the corresponding features of the `instant` crate transitively. This allows the use of `conrod` without panic when targeting WASM.

Note that this is not a breaking change because for all existing users the `instant::Instant` type will just be a type alias for `std::time::Instant`.

This PR is related to #1118.

### DIFF
--- a/conrod_core/Cargo.toml
+++ b/conrod_core/Cargo.toml
@@ -17,6 +17,10 @@ categories = ["gui"]
 [package.metadata.docs.rs]
 all-features = true
 
+[features]
+stdweb = [ "instant/stdweb" ]
+wasm-bindgen = [ "instant/wasm-bindgen" ]
+
 [dependencies]
 conrod_derive = { path = "../conrod_derive", version = "0.63" }
 daggy = "0.5"
@@ -24,3 +28,4 @@ fnv = "1.0"
 num = "0.2"
 pistoncore-input = "0.24"
 rusttype = { version = "0.7", features = ["gpu_cache"] }
+instant = "0.1"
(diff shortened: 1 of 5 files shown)


mitchmindtree:
Thanks @sebcrozet! This looks fine to me.

sebcrozet:
@mitchmindtree Thanks! Would it be possible to publish a new release of `conrod`?

mitchmindtree:
Yes, I'm planning to publish a new version of each of the conrod crates within the next couple of days - I just want to check #1266 first and investigate a bug where the vulkano backend text rendering is appearing distorted after applying some fixes to image layout transitions upstream. I hope that's OK!

sebcrozet:
@mitchmindtree By any chance, do you think a release of `conrod_core` could happen before the end of the month? I'd love to be able to release `kiss3d` with `conrod` support before the end of the month too so that I can communicate about this integration in my next monthly newsletter.

PR #1266, opened 2019-04-12
Fix incorrectly scaled text when there is a non integer dpi_factor

### LINKED ISSUE #1265: The text cursor position is not aligned with actual character positions in TextEdit
image
The error seems to add up moving across a line such that it is very visible at the end of longer lines and seems to be exacerbated by smaller font sizes. Here the cursor should be right after the h but you can see that it is significantly offset.

Imberflur:
fixes #1265
If there is better/more preferred solution for this, I would be happy to implement it.


mitchmindtree:
Just tested this locally and all LGTM - thanks and sorry about the delay!

Issue #955, opened 2017-02-26
Piston window doesn't work with DropDownList


chengchangwu:
I've changed examples/support/mod.rs to the following to test DropDownList.
With `cargo run`,

    cargo run --features="winit glium" --example all_winit_glium --release
    cargo run --features="piston" --example all_piston_window --release

The piston version doesn't work.

```
#![allow(dead_code)]

use conrod;
use std;

#[cfg(feature="glium")] use conrod::backend::glium::glium;


pub const WIN_W: u32 = 600;
pub const WIN_H: u32 = 420;


/// A demonstration of some application state we want to control with a conrod GUI.
pub struct DemoApp {
    selected_idx: Option<usize>,
    colors: Vec<String>,
}


impl DemoApp {

    /// Simple constructor for the `DemoApp`.
    pub fn new(rust_logo: conrod::image::Id) -> Self {
        DemoApp {
            selected_idx: None,
            colors: vec!["BLACK".to_string(), "YELLOW".to_string(), "RED".to_string(), "GREEN".to_string(), "BLUE".to_string()],
        }
    }

}


/// A set of reasonable stylistic defaults that works for the `gui` below.
pub fn theme() -> conrod::Theme {
    use conrod::position::{Align, Direction, Padding, Position, Relative};
    conrod::Theme {
        name: "Demo Theme".to_string(),
        padding: Padding::none(),
        x_position: Position::Relative(Relative::Align(Align::Start), None),
        y_position: Position::Relative(Relative::Direction(Direction::Backwards, 20.0), None),
        background_color: conrod::color::DARK_CHARCOAL,
        shape_color: conrod::color::LIGHT_CHARCOAL,
        border_color: conrod::color::BLACK,
        border_width: 0.0,
        label_color: conrod::color::WHITE,
        font_id: None,
        font_size_large: 26,
        font_size_medium: 18,
        font_size_small: 12,
        widget_styling: std::collections::HashMap::new(),
        mouse_drag_threshold: 0.0,
        double_click_threshold: std::time::Duration::from_millis(500),
    }
}


// Generate a unique `WidgetId` for each widget.
widget_ids! {
    pub struct Ids {
        canvas,
        color_select,
    }
}

    pub fn gui(ui: &mut conrod::UiCell, ids: &Ids, app: &mut DemoApp) {
        use conrod::{widget, Labelable, Positionable, Sizeable, Widget};

        // We can use this `Canvas` as a parent Widget upon which we can place other widgets.
[...]

Issue #1166, opened 2018-04-24
Bug in gfx backend when using crop_kids


Epsylon42:
It seems like the gfx backend incorrectly calculates the area's y coordinate.

For example
Canvas in the middle:
![default](https://user-images.githubusercontent.com/13032641/39199366-fa76efe4-47f1-11e8-8e19-f4786cff7e4c.png)

Canvas moved down:
![default](https://user-images.githubusercontent.com/13032641/39199450-259e5db0-47f2-11e8-90be-7b3904aa4db6.png)
Whoops. I can't see half of it.

Canvas moved up:
![default](https://user-images.githubusercontent.com/13032641/39199518-5586c0da-47f2-11e8-8851-90a9a5731d13.png)
And here's a text area I probably should not see.

I managed to kinda fix it by changing rect_to_gfx_rect [here](https://github.com/PistonDevelopers/conrod/blob/d595b2cfd5049e9297761d2448e20d2b253ccec5/src/backend/gfx.rs#L277) like this:

```rust
let rect_to_gfx_rect = |rect: Rect| {
    let (w, h) = rect.w_h();
    let left = (rect.left() * dpi_factor + half_win_w) as u16;
    let bottom = (-rect.bottom() * dpi_factor - (h * dpi_factor) + half_win_h) as u16;
    let width = (w * dpi_factor) as u16;
    let height = (h * dpi_factor) as u16;
    gfx::Rect {
        x: std::cmp::max(left, 0),
        w: std::cmp::min(width, screen_w as u16),
        y: std::cmp::max(bottom, 0),
        h:  std::cmp::min(height, screen_h as u16),
    }
};
```

This fixes the problem, but now the label above the canvas is invisible. Also this solution doesn't seem right since the glium backend has the same conversion function and it works correctly.

tdaffin:
This looks a lot like a problem I ran into.
I've posted a PR to fix it: https://github.com/PistonDevelopers/conrod/pull/1274

Perhaps you could try using the branch my fix is on with your code to see if it fixes your problem?

You'd use the following two lines in your Cargo.toml:

```
[dependencies]
conrod_core = { git = "https://github.com/tdaffin/conrod", branch = "fix_crop" }
conrod_piston = { git = "https://github.com/tdaffin/conrod", branch = "fix_crop" }
```

Epsylon42:
I'm using gfx backend, so the fix doesn't work for me. As far as I know, the bug is present only there.

[...]

PR #1274, opened 2019-05-06
Minimal fix for list cropping problem

tdaffin:
This fixes a problem where a scrollable canvas gets strangely cropped when inside another scrollable region.
This also applies to List, ListSelect and DropDownList.
To demonstrate the problem I've added a scrollable canvas and a List to the all_piston_window example on this commit:
tdaffin@b84ae50

I think that this will fix https://github.com/PistonDevelopers/conrod/issues/955

### DIFF
--- a/backends/conrod_piston/src/draw.rs
+++ b/backends/conrod_piston/src/draw.rs
@@ -289,8 +289,5 @@ fn crop_context(context: Context, rect: Rect) -> Context {
         }
     }
 
-    // The y coordinate is backwards, min to prevent overflow
-    let y = draw_dim[1] as u32 - (y+h).min(draw_dim[1] as u32);
-
     Context { draw_state: draw_state.scissor([x, y, w, h]), ..context }
 }


tdaffin:
In case it helps, this change essentially rolls back this PR: https://github.com/PistonDevelopers/conrod/pull/1230 
Perhaps this is a platform/window manager dependent problem?
I'm running on Ubuntu 18.04, but I'm not sure what platform @0e4ef622 was running on.

0e4ef622:
I was running Arch Linux with i3wm.

0e4ef622:
This might be related to https://github.com/PistonDevelopers/opengl_graphics/issues/298

0e4ef622:
And this reintroduces the exact bug that the code you're reverting fixes. I checked just to make sure.

0e4ef622:
Hm that's strange, with your fix, the y coordinate for the scissor is backwards in my project but correct with your example, and vice versa without.

tdaffin:
Can you post the code of your project @0e4ef622? If we could get your example and mine into the same layout in the same example it could be useful.

0e4ef622:
[Here's a gist](https://gist.github.com/0e4ef622/9662e5da9ac20bbe0a2a0210d5923a93). You'll need to add an arial.ttf to `src/` or edit the path to the font file.

tdaffin:
Thanks.  I can reproduce your bug using this.
I'm working on making a single executable that exhibits both behaviors...

tdaffin:
I've created a repository which uses the code from @0e4ef622's gist to show both behaviors.
https://github.com/tdaffin/conrod_bug
So far it looks like the main difference is that @0e4ef622's (in gl and gl_rend modules) uses opengl_graphics::GlGraphics whereas mine (in pist and pist_rend modules) uses piston_window::G2d.
I'm going to try to make the two versions use more of the same code while still showing the two different behaviors in an effort to home in on the real cause of the bug.

tdaffin:
I made some more changes to the repro code to try and narrow down the bug some more.
I can't make both windows interactive at once (get a segfault), but I can get a screenshot of the two:
image

0e4ef622:
@tdaffin I think you should move this stuff to a new issue.

tdaffin:
Yes, a new issue would probably be a good idea.
Not sure where to put it though so I'll just jot down my findings so far here...
At this point it looks like the problem is between two different ways to define coordinates for the scissor.  
I'm not 100% sure as yet, but I think that most of piston/conrod and the back ends define coordinates relative to the upper left corner with x going left to right and y going down.
https://github.com/PistonDevelopers/graphics/blob/master/src/draw_state.rs looks like it is used that way, though it only mentions 'in screen space' by way of definition.

However, once the [u32; 4] arrives down inside of the gl_graphics backend in the bind_scissor function in https://github.com/PistonDevelopers/opengl_graphics/blob/master/src/draw_state.rs and is passed to gl::Scissor then it is required to be defined relative to the lower left corner with x going left to right and y going up.  See https://www.khronos.org/opengl/wiki/Scissor_Test

... At least that is what I *think* is going on...

tdaffin:
I've opened a new issue in opengl_graphics (https://github.com/PistonDevelopers/opengl_graphics/issues/302) with a PR to fix it (https://github.com/PistonDevelopers/opengl_graphics/pull/303) and posted a fixed version of the example that references both this PR and the one for conrod: https://github.com/tdaffin/conrod_bug/tree/fix_crop

I suppose it will be necessary to bump the version number on opengl_graphics once that PR is merged so that this PR can require the fixed version of opengl_graphics.

tdaffin:
@0e4ef622 does the fix I made to piston2d-opengl_graphics make this this fix work for you?
It has been merged and published as version 0.60 and appears to make everything work in https://github.com/tdaffin/conrod_bug

0e4ef622:
Yep, the fix works. Nice job narrowing down the bug farther than I did when I first found it.

tdaffin:
This may also fix https://github.com/PistonDevelopers/conrod/issues/1166

mitchmindtree:
Thanks for looking into this @tdaffin and @0e4ef622!

What is the status of this PR? Would you still like to land this?

tdaffin:
Yes, I think this should be merged with an increment to the conrod crate version.
That way those who are already using conrod 0.63 with piston2d-opengl_graphics 0.59 will just need to take 0.60 of piston2d-opengl_graphics to fix any problems when they update conrod.

tdaffin:
FYI: A separate fix for https://github.com/PistonDevelopers/conrod/issues/1166 will be required.

mitchmindtree:
OK thanks @tdaffin, I'm also interested in publishing a new version for the recent `conrod_vulkano` updates so I'll merge this and get to it :+1:

tdaffin:
Thanks.  I've just been testing the new version and all seems well...

PR #1275, opened 2019-05-10
Better triangulation in polygon

alanpoon:
The current triangulation causes all triangles to use the first point. Hence it is not able to render the following image, when the first point is the left point of the arrowhead.
Screenshot from 2019-05-10 09-59-11


conrod_core/src/widget/primitive/shape/polygon.rs
```diff
@@ -292,11 +292,19 @@ impl<I> Iterator for Triangles<I>
...
+        };
+        let prev = match self.points.next() {
+            Some(p) => p,
+            None => return None,
+        };
+        let point = match self.points.next() {
+            Some(p) => p,
+            None => return None,
+        };
+        Some(Triangle([first, prev, point]))
```
tafia:
```rust
Some(Triangle([self.points.next()?, self.points.next()?, self.points.next()?]))
```

conrod_core/src/widget/primitive/shape/polygon.rs
```diff
@@ -232,6 +232,14 @@ impl<I> Widget for Polygon<I>
                     state.update(|state| state.points.truncate(total)),
                 None => (),
             }
+            state.update(|state| state.points.push(state.points[0]));//close polygon
+            let triangles = triangulate(&state.points);
+            let mut points_k = vec![];
+
+            for ta in triangles{
+                points_k.push([state.points[ta][0],state.points[ta][1]]);
+            }
```
tafia:
```rust
let points_k = triangles.into_iter().map(|ta| [state.points[ta][0],state.points[ta][1]]).collect();
```

conrod_core/src/widget/primitive/shape/polygon.rs
```diff
@@ -232,6 +232,14 @@ impl<I> Widget for Polygon<I>
                     state.update(|state| state.points.truncate(total)),
                 None => (),
```
mitchmindtree: Can we add an early return to this `None` branch so that we avoid doing the following `state.update` calls in the case that there has been no change to the widget instance? Calls to `state.update` indicate to conrod that the widget has changed in some way and requires re-drawing, so it's good to minimize these calls whenever possible.

conrod_core/src/widget/primitive/shape/polygon.rs
```diff
@@ -271,18 +279,10 @@ impl<I> Colorable for Polygon<I> {
...
-    let prev = match points.next() {
-        Some(p) => p,
-        None => return None,
-    };
+    let points = points.into_iter();
     Some(Triangles {
-        first: first,
-        prev: prev,
+        first: [0.0,0.0],
+        prev: [0.0,0.0],
```
mitchmindtree: Are these `first` and `prev` fields still used? Could they possibly be removed now that we retrieve three new points each time we yield a triangle? Or are they still used elsewhere?

alanpoon:
I think I will be better of using primitive::shape::triangles, because of the complicated triangulation. I will close the PR.

PR #1280, opened 2019-05-26
Shared examples

tdaffin:
Kind of a speculative PR here, where I'm trying to achieve a few goals:

* Make the shared example code more readable by breaking it up into a bunch of sub modules
* Make it easier to add new examples to the shared examples without having to change any of the backends
* Separate the Ids into a bunch of smaller structs so that simple names like 'button' can be used in each

I'd like to move some of the examples from conrod_glium into shared examples, and I think that this restructuring will make that easier.

I think that it would perhaps also be useful to add something like a non scrolling menu bar at the top of the shared examples.  This menu could have controls that allow sections of the shared examples to be dynamically swapped out.  It could perhaps also allow the sections to be displayed in different contexts like tabs, or in scrolling.  I expect that contextual problems like scissor rectangles could then be more easily reproduced.

### DIFF
--- a/backends/conrod_example_shared/src/button_xy_pad_toggle.rs
+++ b/backends/conrod_example_shared/src/button_xy_pad_toggle.rs
@@ -49,11 +49,11 @@ impl Gui {
     }
 
     /// Returns id of widget that the next Gui should be down_from
-    pub fn update(&self, ui: &mut UiCell, state: &mut GuiState, canvas: widget::Id, rect: &Rect, side: f64) -> widget::Id {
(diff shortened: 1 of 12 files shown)


tdaffin:
I've been continuing development on top of this branch (https://github.com/tdaffin/conrod/tree/move_examples - not ready for PR yet), and as a result I think perhaps calling all the structs that hold ids 'Gui' isn't such a great idea.
I'm not entirely sure what would be a better name -- perhaps something like 'Widgets' ?

PR #1285, opened 2019-06-23
Update some more local helper macro use

csirkeee:
Change some local helper macro use, that was added since my last pull request, according to
https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html

This way, if someone uses `conrod_winit::conversion_fns` without importing in all the helper macros, their code will still compile.


mitchmindtree:
This was definitely an oversight on my behalf, thanks a lot for this!

PR #1287, opened 2019-06-29
Change conrod_vulkano `Image` type to use a more flexible format

### LINKED ISSUE #1286: [conrod_vulkano] Image data should be interpreted as sRGB format
Here's what the image looks like currently in the `all_winit_vulkano` example:

Screenshot from 2019-06-29 21-26-23

Here's what it looks like with the image format changed to `R8G8B8A8Srgb`.

Screenshot from 2019-06-29 21-24-12

This makes sense, as sRGB is generally used as a storage format. Ideally, we'd check for whether or not something is stored as non-linear sRGB or not using the `image` crate somehow, but I'm currently not sure how to do this.

mitchmindtree:
This allows for loading multiple images with different underlying
formats.

Also fixes a bug in the example to use sRGB which closes #1286.

### DIFF
--- a/backends/conrod_vulkano/examples/all_winit_vulkano.rs
+++ b/backends/conrod_vulkano/examples/all_winit_vulkano.rs
@@ -78,7 +78,7 @@ fn main() {
             width: logo_dimensions.0,
             height: logo_dimensions.1,
         },
-        vulkano::format::R8G8B8A8Unorm,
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Travis seems to be failing due to an unrelated bug on the latest nightly, going to merge.

PR #1290, opened 2019-07-02
Update conrod_vulkano to vulkano version 0.13

mitchmindtree:
This fixes a UB related bug. See vulkano-rs/vulkano#1212 for more
details.

Unfortunately the fix for this requires that vulkano `Vertex` types
implement `Default`, meaning this is a breaking change, albeit a slight
one.


mitchmindtree:
Travis just seems to be failing due to a rustdoc bug on nightly, going to merge.

PR #1300, opened 2019-08-18
Update shaderc dep in conrod_vulkano. Update gfx deps. Publish 0.67.

mitchmindtree:
This is kind of a late, emergency release to fix an issue where users of
some Linux distros (Arch, Void, possibly others) cannot currently build
conrod_vulkano due to a bug in the shaderc 0.5.1 build script.
google/shaderc-rs#58.

Edit: Travis was showing a build error on the nightly build of an older
dependency. The gfx deps have been updated as an attempt to fix this.

### DIFF
--- a/backends/conrod_example_shared/Cargo.toml
+++ b/backends/conrod_example_shared/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conrod_example_shared"
-version = "0.66.0"
+version = "0.67.0"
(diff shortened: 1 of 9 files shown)


mitchmindtree:
Seeing as stable and beta are building fine, I'm going to assume this is a bug in the latest rustc nightly. I've opened up an issue to track this at gfx-rs/gfx#2964.

Going begin publishing and merge now. If it turns out to be a gfx bug, we can patch the `conrod_gfx` crate with the necessary fix later on.

PR #1303, opened 2019-09-29
Make delta_xy of Drag events add up to total_delta_xy

bkloster:
With high frame rates, slow mouse movement or high drag thresholds,
Ui::handle_event omits any Drag events whose delta_xy would be below the
threshold. That causes the sum of all delta_xy of sequential Drag events not
to add up to total_delta_xy.

The solution implemented here is to compare total_distance_xy instead of
delta_xy against the threshold. This is admittedly not ideal. A better
solution would be to set a flag when the first movement exceeds the threshold,
then generate Drag events as long as the flag is set. Once the mouse button
is released, reset the flag to stop generating Drag events.

### DIFF
--- a/conrod_core/src/tests/ui.rs
+++ b/conrod_core/src/tests/ui.rs
@@ -106,6 +106,50 @@ fn ui_should_reset_global_input_after_widget_are_set() {
 
 
 #[test]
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Thanks for this @bkloster, and sorry for the huge delay!

PR #1308, opened 2019-10-12
Implement Copy and Paste for TextBox

shanehandley:
This PR implements Copy and Paste control via CTRL + C and CTRL + V.

Note that this introduces a new dependency on [copypasta](https://docs.rs/crate/copypasta/0.6.0) to allow cipboard access.


conrod_core/src/widget/text_edit.rs
```diff
@@ -242,6 +242,7 @@ impl<'a> Widget for TextEdit<'a> {
         let widget::UpdateArgs { id, state, rect, style, ui, .. } = args;
         let TextEdit { text, .. } = self;
         let mut text = std::borrow::Cow::Borrowed(text);
+        let mut clipboard: ClipboardContext = ClipboardContext::new().unwrap();
```
mitchmindtree: It looks like some of the backends for the `ClipboardContext` make some non-trivial calls each time the `ClipboardContext` is created - at a quick glance, macOS seems to make some obj-c runtime calls and does some checks in order to retrieve this context. I wonder if we're better off creating this once on initialisation of the `Ui` and making it available within the widget update via a method on the `UiCell` instead?

Alternatively, rather than constructing this every update, we could instead move it into the copy and paste key press conditions (e.g. L618 and L672) so that we only pay the cost on copy/paste events?

shanehandley: Thanks @mitchmindtree I've changed this to adhere to the latter suggestion. I like the idea of adding it to `Ui` however I was not able to make this work without removing `#[derive(Debug)]` from the Ui and UiCell structs as various parts of Copypasta do not implement this trait, and I was worried about introducing that break.

On the CI front, the `cargo test --verbose` failure error is a bit obscure:

`error: linking with `cc` failed: exit code: 1`

Then

```
= note: /usr/bin/ld: cannot find -lxcb-shape
          /usr/bin/ld: cannot find -lxcb-xfixes
          collect2: error: ld returned 1 exit status
```

I currently don't have a Linux box to test this on, but I will find one and investigate further. Maybe a missing lib somewhere?

Cheers.

shanehandley: CI should be fixed now. It related to https://github.com/aweinstock314/rust-clipboard/issues/67

Unfortunately I've had to add two new packages to the travis packages list.

mitchmindtree:
Going to start merging these @shanehandley so we can put them out in a new release - thanks so much for your work on this!

PR #1311, opened 2019-10-14
Update documentation link in Cargo config as the current URL is a 404.

### LINKED ISSUE #1306: http://docs.piston.rs/conrod/conrod is currently a 404
Other projects linking from http://docs.piston.rs/ look fine.

shanehandley:
Fixes #1306


mitchmindtree:
Maybe we should just switch to `docs.rs/conrod_core` at this point? I don't remember if there's a reason in particular we're using the piston domain - I think it might just be that conrod pre-dates docs.rs.

PR #1317, opened 2019-11-20
[WIP] Update winit and glium to latest

hatoo:
It is very good for me if `conrod_glium` uses latest `glium` and `winit` in my use case.

This PR updates `glium` and `winit` to the latest version.
I've updated hello_world.rs example only but I'll update other examples if this PR is OK.

### DIFF
--- a/backends/conrod_glium/Cargo.toml
+++ b/backends/conrod_glium/Cargo.toml
@@ -17,7 +17,7 @@ path = "./src/lib.rs"
 
 [dependencies]
 conrod_core = { path = "../../conrod_core", version = "0.68" }
-glium = { version = "0.24" }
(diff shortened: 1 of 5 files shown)


mitchmindtree:
Apologies for the late response to this!

Yes I think updating is more than welcome. If you are still interested in landing this, would you mind rebasing the PR onto master, and using the new `conrod_winit::v021` module rather than modifying the original `conrod_winit` macros? This allows to avoid breaking other backends that rely on an older version of winit.

hatoo:
Thanks for replying.

Sorry, Since I almost forgot about this PR, and it seems my work becomes pretty old now, I close this PR and maybe create another PR if I have time.

PR #1321, opened 2019-12-22
Update dependencies

oherrala:
* conrod_derive: Update proc-macro2, quote and syn crates to 1.0 versions.
 * all other crates: Update dependencies which doesn't break tests.


mitchmindtree:
This is great - thanks so much!

PR #1331, opened 2020-02-11
Add a `conrod_wgpu` backend crate

mitchmindtree:
Just about everything should be implemented other than images, resizing,
and perhaps a MSAA resolve in the example. These shouldn't be tricky to
add, I'll do so soon before merging this PR. _Edit: This is all done now._

Everything went very smoothly, other than the alpha channel written by
the fragment shader seemingly being ignored. At first I though something
must be wrong with the way I was sampling alpha values from the glyph
cache, however even if I manually set the alpha to 0.0 (and ignore the
glyph cache texture sampling entirely) the text still renders as white
rectangles. Will look into this further next time I get the chance if
noone beats me to it. I'm sure I'm probably just missing a flag or bool
somewhere.

cc @DavidPartouche @freesig

This also includes an update to `conrod_winit` to support winit 0.21 via a
new `v021` module.

**Transparency bug**

Here's what the output looks like:

Screenshot from 2020-02-12 00-35-47

Screenshot from 2020-02-12 00-37-05

Here's what it should look like:

Screenshot from 2020-02-12 00-22-20

Screenshot from 2020-02-12 00-23-41


mitchmindtree:
The transparency bug has been fixed thanks to a hot tip from the folks in the wgpu matrix channel - I had failed to specify both the color blend and alpha blend descriptions correctly! Transparency now behaves as expected all round.

I've also added the missing image support, fixed the resizing bug and enabled MSAA in the example.

This PR should be ready to go!

PR #1332, opened 2020-02-15
Create a Display trait for the glium backend

AlexCharlton:
In the master branch, `Renderer::fill` takes` display: &glium::Display` as a parameter which itself is a re-export of `glium::backend::glutin::Display`. As such the conrod glium backend depends on glutin.

By moving the methods of `glium::Display` into a trait we remove this hard dependency, allowing other window contexts to be used by the conrod glium backend. The `Display` trait is implemented for `glium::Display` which makes this change transparent to existing applications.

### DIFF
--- a/backends/conrod_glium/src/lib.rs
+++ b/backends/conrod_glium/src/lib.rs
@@ -416,6 +416,25 @@ impl GlyphCache {
     }
 }
 
+pub trait Display {
+    fn opengl_version(&self) -> &glium::Version;
+    fn framebuffer_dimensions(&self) -> (u32, u32);
+    fn hidpi_factor(&self) -> f64;
+}
+
+impl Display for glium::Display {
+    fn opengl_version(&self) -> &glium::Version {
+        self.get_opengl_version()
+    }
+
+    fn framebuffer_dimensions(&self) -> (u32, u32) {
+        self.get_framebuffer_dimensions()
+    }
+
+    fn hidpi_factor(&self) -> f64 {
+        self.gl_window().get_hidpi_factor()
+    }
+}
 
 impl Renderer {
     /// Construct a new empty `Renderer`.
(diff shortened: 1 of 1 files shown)


backends/conrod_glium/src/lib.rs
```diff
@@ -416,6 +416,25 @@ impl GlyphCache {
...
+    fn framebuffer_dimensions(&self) -> (u32, u32);
+    fn hidpi_factor(&self) -> f64;
+}
+
+impl Display for glium::Display {
+    fn opengl_version(&self) -> &glium::Version {
+        self.get_opengl_version()
+    }
+
+    fn framebuffer_dimensions(&self) -> (u32, u32) {
```
AlexCharlton: I don't believe there's any way for this trait to have a method named `get_framebuffer_dimensions` (or `get_opengl_version` or `get_hidpi_factor`) which would have been nice for consistency. If I'm mistaken, I'd love to hear it.

mitchmindtree:
Thanks @AlexCharlton, this LGTM!

Would you mind sharing what you OpenGL context provider you are using if not `glutin`? I'm mostly just curious! I always knew it was theoretically possible to use `glium` without `glutin`, but have not personally seen it happen in the wild.

AlexCharlton:
@mitchmindtree I'm experimenting with wxWidgets (in order to get native menu bars and such). Here's a (messy) example that works on OSX on this patched branch of Conrod: https://gist.github.com/AlexCharlton/e1804966f5f8fc64d1b0b965142ff89b

backends/conrod_glium/src/lib.rs
AlexCharlton: I figured it out!

AlexCharlton: Nope, nevermind.

PR #1340, opened 2020-03-29
Make winit conversion functions generic over UserEvent

icewind1991:
this makes it possible to use custom user event types with the winit
event loop.


mitchmindtree:
Ahh good catch, thanks for this!

PR #1365, opened 2020-07-18
Update to pistoncore-input 1.0.0

bvssvni:


bvssvni:
I got some problems with different versions of cocoa.

bvssvni:
Maybe the problem with cocoa only happens when using `cargo test` at top level?

bvssvni:
Seems to be a problem with the `conrod_wgpu` backend.

bvssvni:
Travis CI runs `cargo test` at top level, so this might be a problem specific to my machine.

bvssvni:
cc @mitchmindtree

bvssvni:
Merging

PR #1367, opened 2020-07-24
Fix scrollbar behaviour with padded canvas

alvinhochun:
This fixes the behaviour of scrollbars on canvas, when the scrollable content excluding padding fits the kids area but overflows when padding is included. Without this fix, the scrollbar shows a full-range handle, while still scrollable, will cause the scrollable content to disappear after flicking the scrollbar handle a few times, due to the scroll offset turning into NaN.

I have also added some checks to prevent the scroll offset from becoming NaN.


mitchmindtree:
Thanks!

PR #1371, opened 2020-07-31
Fix rgb to hsl saturation

### LINKED ISSUE #1370: rgb_to_hsl can return unexpected values
I could be wrong, and I apologize if so, but I think there are two issues with the color `rgb_to_hsl` function.

My understanding was that `Color::Rgba` and `Color::Hsla` should have between 0.0 and 1.0 for all of their values. If that is true, then the hue can often be outside of this range.

The second problem is that there is an edge case missed when calculating `saturation` as a value of 1 for `lightness` will return a NaN. This case can be seen in the Wikipedia page on [HSL and HSV](https://en.wikipedia.org/wiki/HSL_and_HSV#From_RGB).

These can be reproduced with `rgb_to_hsl(0.60, 0.23, 0.89)` and `rgb_to_hsl(1.0, 1.0, 1.0)` respectively returning `(4.77585554, 0.74999994, 0.560000002)` and `(0, NaN, 0)`

Could anyone can confirm this problem is not just me using the function incorrectly?

EDIT: I should also add that the hue seems to be calculated incorrectly too. For instance, purple (128, 0, 128) should return (300 degrees, 100%, 25%) yet `rgb_to_hsl(0.5, 0.0, 0.5)` returns `(5.23598766, 1.0, 0.25)`... multiplying the h by 60 returns the incorrect value of `314.1592596`.

sbauer322:
Closes #1370 

Addresses the NaN problem mentioned in #1370 . It seems like the hue range is expected to be between 0 and 2PI for other parts of the library, so I'm assuming that is correct and left it as is.

Also expanded the doc for the function indicating the expected input and output ranges.

### DIFF
--- a/conrod_core/src/color.rs
+++ b/conrod_core/src/color.rs
@@ -329,7 +329,8 @@ pub fn f32_to_byte(c: f32) -> u8 { (c * 255.0) as u8 }
 
 
 /// Pure function for converting rgb to hsl.
-/// Inputs expected to be between `0.0` and `1.0`.
+/// * Inputs expected to be between `0.0` and `1.0`.
+/// * Outputs `[0.0, 2*PI)` for `h`, `[0.0, 1.0]` for both `s` and `l`
 pub fn rgb_to_hsl(r: f32, g: f32, b: f32) -> (f32, f32, f32) {
     let c_max = r.max(g).max(b);
     let c_min = r.min(g).min(b);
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Thanks so much for this @sbauer322! Eventually, I hope that we can find the time to remove conrod's colour module in favour of using `palette`. That said this fix is great for the meantime!

PR #1373, opened 2020-08-08
Change to round scissor rect before truncating

alvinhochun:
This change rounds the scissor rect before truncating to an integer type.

**Rationale:**

When a fractional hidpi scaling is in use, the code may want to align widgets or graphical elements to the physical device pixel grid. This can be done by `aligned = (coord * hidpi_factor).round() / hidpi_factor` for example, where `coord` is a coordinate relative to the corner (not centre) of the viewport. This is done so that when the coordinate is converted back to the device coordinate by `aligned * hidpi_factor`, the result is an integer or close to an integer.

Unfortunately, due to the nature of floating point maths, the end result is often not an exact integer. Therefore when the scissor code truncates the coordinates, an off-by-one error may occur. Rounding the value before truncation fixes the issue.

**Potential issues:**

This may result in a different behaviour for widgets that does not explicitly align themselves to the pixel grid, even when there is no dpi scaling.

**Other notes:**

This PR changes all included backends except for the Piston backend. (The calculation in the Piston backend is quite different from the others so I'm unsure how to properly change it.)

### DIFF
--- a/backends/conrod_gfx/src/lib.rs
+++ b/backends/conrod_gfx/src/lib.rs
@@ -290,10 +290,10 @@ impl<'a, R: Resources> Renderer<'a, R> {
 
         let rect_to_gfx_rect = |rect: Rect| {
             let (w, h) = rect.w_h();
-            let left = (rect.left() * dpi_factor + half_win_w) as u16;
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Looks good to me, thanks!

PR #1374, opened 2020-08-08
Update glium backend to glium 0.28 with winit 0.23

alvinhochun:
This updates the glium backend to use glium 0.28 with winit 0.23.

The examples are also updated to use the new `EventLoop::run` idiom as the polling method previously used is no longer available. This results in some radical changes in the example codes.

I have checked that all the examples works as expected (except for the tutorial code).

I added the winit 0.22 conversion functions (`v022_conversion_fns`) just forwards to the v021 ones as there doesn't seem to be any breaking changes. The 0.23 ones needs some changes though.

### DIFF
--- a/backends/conrod_glium/Cargo.toml
+++ b/backends/conrod_glium/Cargo.toml
@@ -16,7 +16,7 @@ path = "./src/lib.rs"
 
 [dependencies]
 conrod_core = { path = "../../conrod_core", version = "0.70" }
-glium = "0.24"
+glium = "0.27"
 
 [dev-dependencies]
 conrod_example_shared = { path = "../conrod_example_shared", version = "0.70" }
@@ -25,4 +25,4 @@ find_folder = "0.3.0"
 image = "0.22"
 petgraph = "0.4"
 rand = "0.7"
-winit = "0.19"
+winit = "0.22"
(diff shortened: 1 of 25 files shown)


alvinhochun:
Seems that my attempt to be smart about redrawing failed. On Windows, if part of the window was obscured during last redraw, the obscured part will not get rendered to the framebuffer. Going to fix that by making it less smart...

alvinhochun:
Fixed the redraw issue.

alvinhochun:
I redid the redraw fix with another approach so that redrawing happens from within `Event::RedrawRequested`.

mitchmindtree (approved):

mitchmindtree:
Thanks for all these great PRs @alvinhochun :) Feel free to merge this when you're ready and I'll publish `0.71` of each of the conrod crates.

alvinhochun:
I might bump this up to `winit` 0.23 which just got released a few days ago. It looks like `glium` is preparing to make a new release with it through `glutin` 0.25. It might have to wait a few days.

I'm not sure if I want to update the other backends at the same time though.

alvinhochun:
I'm going to bump the wgpu example to winit 0.23 too.

PR #1385, opened 2020-08-25
Update wgpu to 0.6

### LINKED ISSUE #1381: Upgrade dependencies to wgpu(-core) 0.6.0
wgpu(-core) 0.6.0 have been released recently. The new versions seem to support WASM targets, so it would be cool if conrod's dependencies were updated.

yutannihilation:
Fix #1381

### DIFF
--- a/backends/conrod_wgpu/examples/all_winit_wgpu.rs
+++ b/backends/conrod_wgpu/examples/all_winit_wgpu.rs
@@ -1,6 +1,7 @@
 //! An example demonstrating the use of `conrod_wgpu` alongside `winit`.
 
 use conrod_example_shared::{WIN_H, WIN_W};
+use wgpu::util::DeviceExt;
 use winit::{
     event,
     event_loop::{ControlFlow, EventLoop},
@@ -15,6 +16,9 @@ const MSAA_SAMPLES: u32 = 4;
 fn main() {
     let event_loop = EventLoop::new();
 
+    let backends = wgpu::BackendBit::PRIMARY;
+    let instance = wgpu::Instance::new(backends);
+
     // Create the window and surface.
     #[cfg(not(feature = "gl"))]
     let (window, mut size, surface) = {
@@ -27,25 +31,24 @@ fn main() {
             .build(&event_loop)
             .unwrap();
         let size = window.inner_size();
-        let surface = wgpu::Surface::create(&window);
+        let surface = unsafe { instance.create_surface(&window) };
         (window, size, surface)
     };
 
     // Select an adapter and gpu device.
-    let backends = wgpu::BackendBit::PRIMARY;
     let adapter_opts = wgpu::RequestAdapterOptions {
         power_preference: wgpu::PowerPreference::Default,
         compatible_surface: Some(&surface),
     };
-    let adapter_request = wgpu::Adapter::request(&adapter_opts, backends);
-    let adapter = futures::executor::block_on(adapter_request).unwrap();
-    let extensions = wgpu::Extensions {
-        anisotropic_filtering: false,
-    };
+    let adapter = futures::executor::block_on(instance.request_adapter(&adapter_opts)).unwrap();
     let limits = wgpu::Limits::default();
-    let device_desc = wgpu::DeviceDescriptor { extensions, limits };
-    let device_request = adapter.request_device(&device_desc);
-    let (device, mut queue) = futures::executor::block_on(device_request);
+    let device_desc = wgpu::DeviceDescriptor {
+        features: wgpu::Features::empty(),
+        limits,
+        shader_validation: true,
+    };
+    let device_request = adapter.request_device(&device_desc, None);
+    let (device, mut queue) = futures::executor::block_on(device_request).unwrap();
 
     // Create the swapchain.
     let format = wgpu::TextureFormat::Bgra8UnormSrgb;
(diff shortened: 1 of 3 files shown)


yutannihilation:
`cargo test` seems to pass, but I'm seeing this error with the example `all_winit_wgpu`. But, the example fails for me even on master, so I'm not sure if this is the problem of my environment (GNOME3 + Wayland, Intel UHD).

```
thread 'main' panicked at 'the given sampler is/is not a comparison sampler, while the layout type indicates otherwise', /home/yutani/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.6.0/src/backend/direct.rs:1323:35
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
VUID-vkDestroyInstance-instance-00629(ERROR / SPEC): msgNum: -1958900200 - Validation Error: [ VUID-vkDestroyInstance-instance-00629 ] Object 0: handle = 0x560429642ee0, type = VK_OBJECT_TYPE_INSTANCE; Object 1: handle = 0x20000000002, type = VK_OBJECT_TYPE_SURFACE_KHR; | MessageID = 0x8b3d8e18 | OBJ ERROR : For VkInstance 0x560429642ee0[], VkSurfaceKHR 0x20000000002[] has not been destroyed. The Vulkan spec states: All child objects created using instance must have been destroyed prior to destroying instance (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkDestroyInstance-instance-00629)
    Objects: 2
        [0] 0x560429642ee0, type: 1, name: NULL
        [1] 0x20000000002, type: 1000000000, name: NULL
```

backends/conrod_wgpu/src/lib.rs
```diff
@@ -563,11 +567,13 @@ fn bind_group_layout(
             component_type: GLYPH_TEX_COMPONENT_TY,
             dimension: wgpu::TextureViewDimension::D2,
         },
+        count: None,
     };
     let sampler_binding = wgpu::BindGroupLayoutEntry {
         binding: 1,
         visibility: wgpu::ShaderStage::FRAGMENT,
-        ty: wgpu::BindingType::Sampler { comparison: false },
+        ty: wgpu::BindingType::Sampler { comparison: true },
```
yutannihilation, line 575: This change is needed because `sampler_desc()` returns a sampler that does comparison (`compare: Some(wgpu::CompareFunction::Always)`)

backends/conrod_wgpu/examples/all_winit_wgpu.rs
```diff
@@ -306,9 +302,9 @@ fn create_logo_texture(
     let cmd_encoder_desc = wgpu::CommandEncoderDescriptor {
         label: Some("conrod_upload_image_command_encoder"),
     };
-    let mut encoder = device.create_command_encoder(&cmd_encoder_desc);
-    encoder.copy_buffer_to_texture(buffer_copy_view, texture_copy_view, extent);
-    queue.submit(&[encoder.finish()]);
+    let encoder = device.create_command_encoder(&cmd_encoder_desc);
+    queue.write_texture(texture_copy_view, data, data_layout, extent);
```
yutannihilation, line 306: This change is needed because wgpu now is a bit stricter about the data alignment so we'll see this error otherwise:

```
thread 'main' panicked at 'bytes per row does not respect `COPY_BYTES_PER_ROW_ALIGNMENT`', /home/yutani/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.6.0/src/backend/direct.rs:1323:35
```

yutannihilation (commented):
Figured out the cause of errors. I think this is ready now.

alvinhochun:
I can't really comment on the changes but I can confirm that the `all_winit_wgpu` example runs fine on Windows (as opposed to crashing on 9 out of 10 launches with master).

The example itself has some issues with redraws but unrelated to this change. I will make a separate PR for that.

yutannihilation:
Oh, thanks for quick confirmation! I see.

backends/conrod_wgpu/src/lib.rs
```diff
@@ -784,6 +792,20 @@ fn texture_format_to_component_type(format: wgpu::TextureFormat) -> wgpu::Textur
         | wgpu::TextureFormat::Rgb10a2Unorm
         | wgpu::TextureFormat::Depth32Float
         | wgpu::TextureFormat::Depth24Plus
-        | wgpu::TextureFormat::Depth24PlusStencil8 => wgpu::TextureComponentType::Float,
+        | wgpu::TextureFormat::Depth24PlusStencil8
```
alvinhochun, line 795: Do you think it makes sense to handle the TODO comment from above instead of adding to this?

```
// TODO: Remove this when this lands https://github.com/gfx-rs/wgpu/pull/628
```

yutannihilation: Ah, yes. I didn't notice it, sorry.

backends/conrod_wgpu/src/lib.rs
```diff
@@ -537,8 +539,9 @@ fn default_image_tex_desc() -> wgpu::TextureDescriptor<'static> {
     }
 }
 
-fn sampler_desc() -> wgpu::SamplerDescriptor {
+fn sampler_desc<'a>() -> wgpu::SamplerDescriptor<'a> {
```
alvinhochun, line 542: Just a nitpick, but I think this function could just return `wgpu::SamplerDescriptor<'static>` so you don't need the lifetime parameter `'a`.

yutannihilation: Thanks so much, probably I don't understand how to mark lifetimes properly... (I'm still a newbie to Rust)

alvinhochun:
I am considering merging this PR in a week if @mitchmindtree don't have the time to merge, because this makes the wgpu example work on Windows.

Issue #1350, opened 2020-05-10
error: failed to run custom build command for `shaderc-sys v0.6.2`


austern:
Cloned the repository with "git clone https://github.com/PistonDevelopers/conrod.git". Current HEAD commit ID is 0c61e49afba22964a9479c371c7aaca3ab96ca67.

I switched to conrod directory and ran "cargo build --release". The build failed with the following error:

error: failed to run custom build command for `shaderc-sys v0.6.2`

Caused by:
  process didn't exit successfully: `/Users/austern/src/conrod/target/debug/build/shaderc-sys-f341b3d5cd06cebe/build-script-build` (exit code: 101)
--- stdout
cargo:warning=System installed library not found.  Falling back to build from source

--- stderr
thread 'main' panicked at '

couldn't find required command: "cmake"

', /Users/austern/.cargo/registry/src/github.com-1ecc6299db9ec823/shaderc-sys-0.6.2/build/cmd_finder.rs:50:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

warning: build failed, waiting for other jobs to finish...
error: build failed

My system is macOS Mojave 10.14.6. I'm using Rust version 1.43.1.

alvinhochun:
Running `cargo build` on the top-level workspace builds all the projects, which is probably not what you want. FWIW `conrod_vulkano` requires some external dependencies to build (probably including cmake) but you can ignore it for most cases.

If you want to run an example in release build, just run `cargo run --example all_winit_glium --release`.

Filius-Patris:
Sure, but [the guide suggests otherwhise](https://docs.rs/conrod_core/0.70.0/conrod_core/guide/chapter_2/index.html#running-the-conrod-examples).

Your argument is fair, so I'll recommend updating the docs?
I'll just setup a PR real quick to continue this discussion over there.

alvinhochun:
Sure. If you're making a PR for this, just delete the whole steps 3 and 4 as they are really not needed.

A-Mackey:
I was able to fix this issue by installing the Vulkan SDK to my computer then restarting. The docs for shaderc mentions that it uses these SDK's so make sure you have one installed onto your computer.

artayasa:
> The docs for shaderc mentions that it uses these SDK's

This need to instal shaderc sir?? I still confiusd and too hard instal it on win11 always error

PR #1386, opened 2020-08-25
Remove 'build project' instruction from guide

Filius-Patris:
### What?
This PR removes the step to run `cargo build --release` on the main project.

### Why?

The guide should be written for newcomers. Newcomers would usually want to check out
the examples to decide for or against this framework. They should not have to care if the
project builds or not (issue trackers are for that).

So I removed step 3 because it is not needed for running the examples. I also changed the
wording of the introduction of the section, because the goal is not to test if everything works
correctly, but to view the examples.

---
See also the discussion on Issue #1350.

### DIFF
--- a/conrod_core/src/guide/chapter_2.rs
+++ b/conrod_core/src/guide/chapter_2.rs
@@ -29,7 +29,7 @@ with the above steps.
 
 ## Running the Conrod Examples
 
-We can test that everything is working by cloning the github repository and running the examples.
+You can view the examples by cloning the github repository and running the examples.
 First, open up the command line on your system and follow these steps:
 
 1. Clone the repo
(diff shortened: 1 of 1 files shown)


Filius-Patris:
Oh, @alvinhochun, you said former step 4 (now step 3), the `cd` command isn't needed either?

alvinhochun:
Well, running examples directly from the workspace root does work for me, but I couldn't find if it is a recent addition to Cargo or not.

Perhaps it would be better to also specify the package, e.g. `cargo run --release -p conrod_glium --example all_winit_glium` but doesn't seem to be required for now, at least there aren't any duplicated example names. I think just running them from the workspace root is better than `cd`-ing into `backends/conrod_glium` as you can't run examples from other backends (e.g. `all_winit_wgpu`) in there.

Filius-Patris:
Can confirm. I'll update the PR later this evening (CET).

Filius-Patris:
There isn't a command to list all packages, is there?

I mean you can just look them up in the `Cargo.toml` but that's a bit messy to describe. Should I list them all, or is this likely to change?

alvinhochun:
Perhaps just don't add the package flag for now...

alvinhochun:
Looks good to me, thanks!

PR #1389, opened 2020-08-27
Fix all_winit_wgpu example missing some redraws

alvinhochun:
This takes the same approach as the glium example in #1374.

Depends on #1385


mitchmindtree (approved):

mitchmindtree:
Ahh it looks like this was addressed in #1385?

alvinhochun:
> 
> 
> Ahh it looks like this was addressed in #1385?

This PR is really about the last commit which is not in #1385. I was going to rebase this PR after merging #1385, but then I changed the approach of handling redraw in #1374 (which this PR was based on) and I still need to update this PR to use the same approach. I was occupied with other stuff so I decided to delay this, probably after merging #1374.

PR #1401, opened 2020-12-09
Update examples in guide 3 to work with v0.61

BartDeWaal:
I was running through the guide and ran into some  issues, so I thought I would edit the fixes into the document


alvinhochun:
Hi, thanks for contributing! However, 0.61 is still quite dated. It would make sense to go straight to `conrod_core` 0.71.

PR #1410, opened 2021-03-21
[WIP] Update wgpu to 0.7

yutannihilation:
This pull requests update wgpu to 0.7, which was released on Feb 2nd, 2021. 

Most of the changes are straightforward, but I see several errors after update, probably due to the stricter validation. Especially, I don't understand what's wrong with this error. `BindGroupLayout`'s `wgpu::BindingType::Sampler`'s `comparison` is true and actual `SamplerDescriptor` has `compare: Some(wgpu::CompareFunction::Always)`...

```
Caused by:
    In Device::create_render_pipeline
      note: label = `conrod_render_pipeline_descriptor`
    error in stage FRAGMENT
    error matching global binding at index 1 in group 0 against the pipeline layout: comparison flag doesn't match the shader
```

### DIFF
--- a/backends/conrod_wgpu/Cargo.toml
+++ b/backends/conrod_wgpu/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 
 [dependencies]
 conrod_core = { path = "../../conrod_core", version = "0.71" }
-wgpu = "0.6"
+wgpu = "0.7"
 
 [dev-dependencies]
 conrod_example_shared = { path = "../conrod_example_shared", version = "0.71" }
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Thanks so much for the PR @yutannihilation!

I think you're right - I think this is the new wgpu validation pointing out an error with our existing use of the comparison function. From [the wgpu docs](https://docs.rs/wgpu/0.7.1/wgpu/enum.CompareFunction.html):

> Comparison function used for depth and stencil operations.

In our case, our pipeline doesn't even have a depth or stencil attachment as conrod is all 2D. Perhaps declaring a comparison function here was always meaningless?

If I remove the comparison function from the `SamplerDescriptor`, and set the associated `comparison` field to false in the bind group layout, it seems to work nicely!

I'll try updating nannou while testing this branch to make sure we're not missing anything before publishing. If all looks OK, I'll open a PR here based on your branch that removes the comparison function and rebases onto master.

yutannihilation:
Wow, thanks for investigating! I think it's best that you'll take this over as I have little knowledge about how conrod works (I just blindly tweaked code to satisfy the compiler). I'm happy to see nannou is getting updated to use wgpu 0.7 at last! :+1:

Issue #1409, opened 2021-02-09
Build failure on rasberry pi.


thesnout:
I see this error after following the instructions in chapter 2:

pi@hovind:~/conrod/conrod $ cargo build
   Compiling shaderc-sys v0.6.3
   Compiling naga v0.2.0
   Compiling futures-executor v0.3.12
   Compiling rendy-core v0.5.1
   Compiling wayland-client v0.23.6
   Compiling vulkano v0.16.0
   Compiling futures v0.3.12
   Compiling wayland-protocols v0.28.3
error[E0609]: no field `shaderf3264` on type `vk::PhysicalDeviceFeatures`
   --> /home/pi/.cargo/registry/src/github.com-1285ae84e5963aae/vulkano-0.16.0/src/features.rs:164:21
    |
164 |     shader_f3264 => shaderf3264,
    |                     ^^^^^^^^^^^ help: a field with a similar name exists: `shaderInt64`

error[E0560]: struct `vk::PhysicalDeviceFeatures` has no field named `shaderf3264`
   --> /home/pi/.cargo/registry/src/github.com-1285ae84e5963aae/vulkano-0.16.0/src/features.rs:164:21
    |
164 |     shader_f3264 => shaderf3264,
    |                     ^^^^^^^^^^^ `vk::PhysicalDeviceFeatures` does not have this field
    |
    = note: available fields are: `robustBufferAccess`, `fullDrawIndexUint32`, `imageCubeArray`, `independentBlend`, `geometryShader` ... and 50 others

error[E0308]: mismatched types
    --> /home/pi/.cargo/registry/src/github.com-1285ae84e5963aae/vulkano-0.16.0/src/command_buffer/sys.rs:1303:38
     |
1303 |         vk.CmdSetBlendConstants(cmd, constants); // TODO: correct to pass array?
     |                                      ^^^^^^^^^
     |                                      |
     |                                      expected `&[f32; 4]`, found array `[f32; 4]`
     |                                      help: consider borrowing here: `&constants`

   Compiling wayland-cursor v0.28.3
   Compiling wayland-egl v0.28.3
   Compiling rendy-resource v0.5.1
   Compiling rendy-command v0.5.1
error: aborting due to 3 previous errors

Some errors have detailed explanations: E0308, E0560, E0609.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `vulkano`.

pi@hovind:~/conrod/conrod $ rustc --version
[...]

PR #1411, opened 2021-04-07
Update to rusttype 0.8.3

mitchmindtree:
Also removes some unnecessary buffering of positioned glyphs.

*Edit: For context on the following, this PR originally tried updating to 0.9.2*.

Landing this is blocked on a patch that exists on rusttype master but has not yet been released:

https://gitlab.redox-os.org/redox-os/rusttype/-/merge_requests/158

Unfortunately, rusttype is currently lacking a maintainer and as a result it's hard to say when/if we'll be able to land this:

https://gitlab.redox-os.org/redox-os/rusttype/-/issues/148

### DIFF
--- a/backends/conrod_gfx/src/lib.rs
+++ b/backends/conrod_gfx/src/lib.rs
@@ -11,7 +11,7 @@ use gfx::{
 
 use conrod_core::{
     color, image, render,
-    text::{rt, GlyphCache},
+    text::{self, rt, GlyphCache},
     Rect, Scalar,
 };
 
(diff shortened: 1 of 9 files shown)


mitchmindtree:
It looks like we can still get away with using `0.8.3` - it seems that whatever upstream breakage I was running into might have been yanked from crates.io in the meantime.

I think we're probably better off sticking to 0.8.3 for now, at least until either rusttype gets a new maintainer or someone is motivated enough to switch over to the `ab_glyph` spiritual-successor crate.

The vulkano CI breakage is unrelated (see #1409).

PR #1412, opened 2021-04-14
Make backend testing more granular in CI

mitchmindtree:
This removes the old `cargo check examples` job in favour of adding
a unique job for each backend. This should make it easier to spot
backend-specific issues, and will allow to better keep track of known
problems with specific backends without accidentally letting new
problems slide in.


mitchmindtree:
`conrod_vulkano` is known to be failing at the moment (see #1409) - merging this as all other checks a passing.

PR #1416, opened 2021-05-13
Bump webgpu to 0.8

AlexEne:
I've upgraded webgpu. I can't run all the tests as I'm on windows and don't have `ninja` installed. Hopefully if there's anything wrong the CI will pick it up.

I did run the webgpu backend tests and those were fine.

### DIFF
--- a/backends/conrod_vulkano/Cargo.toml
+++ b/backends/conrod_vulkano/Cargo.toml
@@ -18,8 +18,8 @@ edition = "2018"
 [dependencies]
 #conrod_core = { path = "../../conrod_core", version = "0.71" }
 conrod_core = "0.71"
-vulkano = "0.17"
-vulkano-shaders = "0.17"
+vulkano = "0.22"
+vulkano-shaders = "0.22"
 
 [dev-dependencies]
 #conrod_example_shared = { path = "../conrod_example_shared", version = "0.71" }
@@ -28,5 +28,5 @@ conrod_example_shared = "0.71"
 conrod_winit = "0.71"
 find_folder = "0.3"
 image = "0.22"
-vulkano-win = "0.17"
+vulkano-win = "0.22"
 winit = "0.19"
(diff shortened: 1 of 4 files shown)


AlexEne:
I noticed the fmt failiure. Checking what vulkano version works.

AlexEne:
So in order to fix the build due to the change [here](https://github.com/vulkano-rs/vulkano/blob/8c0630d3c86c82581991d599489a69d83f89c97c/CHANGELOG_VK_SYS.md#version-053-2020-12-26)
> Rename shaderf3264 to shaderFloat64
I need to upgrade to vulkano at at least `0.20` from the current `0.16`. This will take a while and I think I may need some help as I'm not too familiar with vulkan, so if I can cope with some WGPU-RS changes, this is a bit complicated and will take me some time to figure out as a bunch of things have happened in Vulkan in the meantime.

mitchmindtree:
Hey @AlexEne don't stress about the vulkano backend if you don't have the time! While it would be nice to have it out of date, it was broken before this PR and isn't related - happy to leave that for a future fix. See #1409 for some context.

Going to take a look at this now!

mitchmindtree:
When attempting to run the `all_winit_wgpu` example (NixOS, GNOME Wayland, Intel GPU), I'm running into the following validation error during shader module creation:

```
wgpu error: Validation Error

Caused by:
    In Device::create_shader_module
      note: label = `shaders/frag.spv`
    Function [1] 'main' is invalid
    Required uniformity of control flow for IMPLICIT_LEVEL in [20] is not fulfilled because of Expression([1])


thread 'main' panicked at 'Handling wgpu errors as fatal by default', /home/mindtree/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.8.1/src/backend/direct.rs:1955:5
```

Looking at the diff, I think it is unlikely this is to do with anything you have changed, as the error is related to validation of the SPIR-V shader which hasn't changed in a while. Likely, this is one of the new validation checks picking up something wrong with old code. Will have a look

mitchmindtree:
It's a bit of a strange one - I've opened an issue here https://github.com/gfx-rs/wgpu-rs/issues/912.

PR #1428, opened 2021-06-23
Fix warning

muzudho:
warning: anonymous parameters are deprecated and will be removed in the next edition.

### DIFF
--- a/conrod_core/src/position/mod.rs
+++ b/conrod_core/src/position/mod.rs
@@ -145,10 +145,10 @@ pub enum Dimension {
 /// Thus, **Positionable** can be implemented for *all* types that implement **Widget**.
 pub trait Positionable: Sized {
     /// Build with the given **Position** along the *x* axis.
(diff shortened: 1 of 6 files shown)


backends/conrod_gfx/src/lib.rs
```diff
@@ -815,14 +815,6 @@ impl From<gfx::PipelineStateError<String>> for RendererCreationError {
     }
 }
 
-impl std::error::Error for RendererCreationError {
-    fn description(&self) -> &str {
-        match *self {
-            RendererCreationError::PipelineState(ref e) => std::error::Error::description(e),
-        }
-    }
-}
```
mitchmindtree, line 824: Can we leave the `std::error::Error` implementations and only remove the inner `description` methods instead? Otherwise I think we might break some downstream crates that rely on these error types implementing the `std::error::Error` trait.

mitchmindtree (commented):
Thanks for this! Just one small change/question then I think this is good to land.

PR #1429, opened 2021-07-09
fix for  "panic message is not a string literal".

lubosmudrak:
Fixed "panic message is not a string literal".
format!() macro is not needed, removed for compatibility with Rust 2021 edition.


lubosmudrak:
Closing the pull request because of no response since 2021.

PR #1432, opened 2021-09-01
add editable text widgets into conrod examples

lubosmudrak:
This pull request adds TextEdit and TextBox widgets to the conrod examples. Purpose of this change is to expose inconsistency of handling text input widgets under some backends(for example piston_window backend). 

I am hoping to write a bug report based on this soon.

### DIFF
--- a/backends/conrod_example_shared/src/lib.rs
+++ b/backends/conrod_example_shared/src/lib.rs
@@ -95,13 +95,19 @@ widget_ids! {
         dialer_title,
         number_dialer,
         plot_path,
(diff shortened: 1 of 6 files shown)


lubosmudrak:
Sorry for inconvenience, but is there a problem with rust-fmt check or are there some other issues with this pull request? Im new to commiting to public repositories.
I would like to see this implemented in main to expose bug in piston backend to public.

lubosmudrak:
Closing the pull request because of no response since 2021.

PR #1439, opened 2021-10-08
port to vulkano 0.26

t18b219k:
Vulkano 0.25+0.26 was big change therefore some change has made to DrawCommand

### DIFF
--- a/backends/conrod_vulkano/src/lib.rs
+++ b/backends/conrod_vulkano/src/lib.rs
@@ -9,39 +9,31 @@ use std::sync::Arc;
 use conrod_core::mesh::{self, Mesh};
 use conrod_core::text::rt;
 use conrod_core::{image, render, Rect, Scalar};
-
-use vulkano::buffer::cpu_pool::CpuBufferPool;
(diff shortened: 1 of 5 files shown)


backends/conrod_vulkano/src/lib.rs
```diff
@@ -485,15 +467,25 @@ impl Renderer {
...
-                            let desc_image = Arc::new(
-                                tex_descs
-                                    .next()
-                                    .add_sampled_image(
-                                        ImageView::new(image.image_access.clone()).unwrap(),
-                                        self.sampler.clone(),
-                                    )?
-                                    .build()?,
-                            );
+                            // ??????????? is there any reason to use self.tex_descs
```
mitchmindtree, line 470: It's been so long enough that I don't recall what's going on here - there's a chance this code is old enough that `FixedSizeDescriptorSetsPool` was the only option at that time, and it's slowly been adapted ever since.

mitchmindtree (approved):
Looks good to me, thanks!

PR #1451, opened 2022-01-07
Add Sdl2 backend

TonalidadeHidrica:
This is the Take 3 of providing SDL2 backend for cornod.

### I didn't use OpenGL

When it comes to using SDL2, there are a big choice: to use OpenGL or not to use.  In the prior art,

* #996 authored by @jonas-schievink does not rely on OpenGL (consequently not on `cornod_gfx` etc) but used only the feature of "pure SDL2" (including `SDL2_gfx`).
* #1055 and #1084 authored by @elrnv does use OpenGL. 

In this PR, I took the course of *not* using it, because I wanted to keep the dependencies minimal.

I checked that all the features of `DemoApp` works, but feel free to pointing out anything I am missing.

### Credits go to:
* @elrnv for event conversion code.  I copy-pasted all your work (but included as the author when committing that part!).
* @jonas-schievink for how to render text.  Although I refined the implementation and adopt newer APIs, the work could have not been done without your code.

### About Rust version

Since I could not find the information of MSRV, I developed `conrod_sdl2` crate with `1.57.0` and Rust 2021 Edition.  If I should use older ones, I would update the code.

### About the addition to the dependency

I added `thiserror` to minimize the boilerplates on error handling, and `smallvec` for a iterator-friendly container of up-to-two events.  If this matters, let me know and I'll think the other way.

### Notes on graphics

When you run `all_sdl2`, you will see the thin diagonal lines as well as horizontal and vertical noize-like lines.   These are probably due to overlapping extremely-thin transparent triangles.  Unlike GL families, triangles in SDL seems not be cut out for forming  complex polygons from many triangles.  I think this is inevitable, as long as we use pure SDL2.
image


elrnv:
Nice work! I'm happy you found the old PRs useful, and thank you for the mention :)

